PR #6, opened 2018-08-31
Add an autoconnecting Layertree

marcjansen:
Hat-tip to @chrismayer for the awesome `mapready`-event idea.

### DIFF
--- a/app/view/LayerTree.js
+++ b/app/view/LayerTree.js
@@ -1,0 +1,68 @@
+/**
+ * A Layertree which autoconnects to the map of the application.
+ */
+Ext.define('CpsiMapview.view.LayerTree', {
+    extend: 'Ext.tree.Panel',
+    xtype: 'layertree',
+    requires: [
+        'BasiGX.util.Map',
+        'GeoExt.data.store.LayersTree'
+    ],
+
+    // So that instanciation works without errors, might be changed during
+    // instanciation of the LayerTree.
+    store: {},
+    rootVisible: false,
+    viewConfig: {
+        plugins: { ptype: 'treeviewdragdrop' }
+    },
+
+    /**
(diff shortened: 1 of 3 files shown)


app/view/LayerTree.js
```diff
@@ -0,0 +1,68 @@
...
+        if (map) {
+            cfg.store = me.makeLayerStore(map);
+        } else {
+            CpsiMapview.getApplication().on('mapready', me.autoConnectToMap);
+        }
+        me.callParent([cfg]);
+    },
+
+    /**
+     * Gueses the mapcomponent and assigns the appropriate layers store, if one
```
chrismayer: Typo `Guesses`?

app/view/LayerTree.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * A Layertree which autoconnects to the map of the application.
+ */
+Ext.define('CpsiMapview.view.LayerTree', {
+    extend: 'Ext.tree.Panel',
+    xtype: 'layertree',
```
chrismayer: Please use the prefix `cmv_` for the xtype.

app/view/LayerTree.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * A Layertree which autoconnects to the map of the application.
+ */
+Ext.define('CpsiMapview.view.LayerTree', {
```
chrismayer: Should  we place this in a separate folder/package?

marcjansen: I wasn't sure. right now we only have on tree…

app/view/LayerTree.js
marcjansen: Will do.

app/view/LayerTree.js
marcjansen: Will change, thx

PR #7, opened 2018-08-31
Implement LayerFactory

chrismayer:
This adds a LayerFactory singleton to create OL layers from a JSON configuration. Supported layers at the moment:

  - OSM
  - WMS
  - ESRI REST

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -1,0 +1,184 @@
+/**
+ * [extend description]
+ * @type {String}
+ */
+Ext.define('CpsiMapview.factory.Layer', {
+    alternateClassName: 'LayerFactory',
+    requires: [],
+
+    singleton: true,
+
+    /**
+     * Creates an OpenLayer layer object from a JSON config
+     *
+     * @param  {Object} layerConf  The configuration object
+     * @return {ol.layer.Base} OL layer object
+     */
+    createLayer: function(layerConf) {
+
+        var layerType = layerConf.layerType;
+        var mapLayer;
+
+        switch (layerType) {
+        case 'wms':
+            mapLayer = LayerFactory.createWms(layerConf);
+            break;
+        case 'wfs':
+            mapLayer = LayerFactory.createWFS(layerConf);
+            break;
+        case 'esrirest':
+            mapLayer = LayerFactory.createEsriRest(layerConf);
+            break;
+        case 'osm':
+            mapLayer = LayerFactory.createOsm(layerConf);
+            break;
+        case 'empty':
+            mapLayer = LayerFactory.createEmptyLayer(layerConf);
+            break;
+        case 'bing_aerial':
+            mapLayer = LayerFactory.createBing(layerConf, 'Aerial');
+            break;
+        case 'google_roadmap':
+            mapLayer = LayerFactory.createGoogle(layerConf, 'google.maps.MapTypeId.ROADMAP');
+            break;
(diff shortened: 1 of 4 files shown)


app/factory/Layer.js
```diff
@@ -0,0 +1,184 @@
...
+
+        var layerType = layerConf.layerType;
+        var mapLayer;
+
+        switch (layerType) {
+        case 'wms':
+            mapLayer = LayerFactory.createWms(layerConf);
+            break;
+        case 'wfs':
+            mapLayer = LayerFactory.createWFS(layerConf);
```
marcjansen: We should create an issue wrt to handling abbreviations and casing consistently. Non-blocking of course.

app/factory/Layer.js
```diff
@@ -0,0 +1,184 @@
...
+            break;
+        case 'google_terrain':
+            mapLayer = LayerFactory.createGoogle(layerConf, 'google.maps.MapTypeId.TERRAIN');
+            break;
+        case 'google_hybrid':
+            mapLayer = LayerFactory.createGoogle(layerConf, 'google.maps.MapTypeId.HYBRID');
+            break;
+        case 'google_satellite':
+            mapLayer = LayerFactory.createGoogle(layerConf, 'google.maps.MapTypeId.SATELLITE');
+            break;
```
marcjansen: We either need a 3rd party plugin for these or remove them altogether. Lets create an issue

app/factory/Layer.js
```diff
@@ -0,0 +1,184 @@
...
+            Ext.log.warn('Layer type not known');
+            //do nothing, and return empty layer
+        }
+
+        return mapLayer;
+    },
+
+    createEmptyLayer: function(layerConf) {
+        Ext.log.info('Not implemented yet', layerConf);
+    },
```
marcjansen: We should implement this, so more complex configs worl out of the box.

app/factory/Layer.js
```diff
@@ -0,0 +1,184 @@
...
+            break;
+        case 'arcgisrest':
+            mapLayer = LayerFactory.createArcGISRest(layerConf);
+            break;
+        case 'switchlayer':
+            mapLayer = LayerFactory.createSwitchLayer(layerConf);
+            break;
+        default:
+            Ext.log.warn('Layer type not known');
+            //do nothing, and return empty layer
```
marcjansen: Call to createEmptyLayer missing

app/factory/Layer.js
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * Creates an OGC WMS layer
+     *
+     * @param  {Object} layerConf The configuration object for this layer
+     * @return {ol.layer.Tile}    WMS layer
+     */
+    createWms: function(layerConf) {
+        return new ol.layer.Tile({
+            name: layerConf.text,
+            source: new ol.source.TileWMS({
```
marcjansen: We need a nissue here wrt to having this configurable probably.

app/factory/Layer.js
chrismayer: Agree, this was transferred by some legacy code.

app/factory/Layer.js
chrismayer: See #8

app/factory/Layer.js
chrismayer: They should because the addition of the layers checks if we have valid layer instances

PR #13, opened 2018-09-01
Give Map component a semantic xtype

marcjansen:
Title says it all


marcjansen:
Cheers for the review.

PR #14, opened 2018-09-01
Add a mapfooter skeleton

### LINKED ISSUE #17: Add map footer component
We need a footer component for the map as container for several tools:

image

marcjansen:
This adds a very basic map footer component which holds a resolution combo for now.

This combo shows an odd behaviour on first rendering, the prefix `1 :` is missing, upon zoom changes in the map it updates fine and displays correctly. I'll open a ticket for this, and think this shouldn't block this from being merged.

Fixes #17


chrismayer:
Thanks :+1:

Issue #28, opened 2018-09-13
Open an ExtJS Form by clicking on a Vector Feature


geographika:
A single click should open a new instance of the form. 
The form will be contained in a separate project.

chrismayer:
>A single click should open a new instance of the form.
The form will be contained in a separate project.

@geographika: The code, how to achieve this, is documented in #30. Please close this issue when you are happy with that.

geographika:
After a few false starts and confusion about which controller was the default scope for an event, I have wired this up in an extended PL Map class. 

The extended Map can be seen at: https://github.com/compassinformatics/publiclighting/blob/master/app/view/main/Map.js

This wires up the event so it can be handled by an extended Map Controller class:

```
listeners: {
    'cmv-mapclick': 'onMapClick'
}
```
https://github.com/compassinformatics/publiclighting/blob/master/app/controller/MapController.js

I think this fits in with the MVVM Sencha architecture - a review of this approach would be appreciated.

chrismayer:
The event-binding looks good to me. That's the way it should be done.

A good start to read about ViewController binding is https://docs.sencha.com/extjs/6.2.0/guides/application_architecture/view_controllers.html

Issue #27, opened 2018-09-13
Add Custom Tooltips for Vector Layers


geographika:
To mimic: https://w08-mapserver.compass.ie/mapview-examples/examples/customtooltips.html

Example code: https://w08-mapserver.compass.ie/mapview-examples/examples/customtooltips.js

Source code: https://github.com/compassinformatics/mapview/blob/master/src/gis/ToolTips.js

Tooltips configuration is loaded from the layer config file. Original tooltips were created using a custom OL2 class. 

![image](https://user-images.githubusercontent.com/490840/45496431-b75a3880-b775-11e8-8744-b4e89655adc9.png)

In mapview tooltips are configured with the tooltipsconfig object. See https://w08-mapserver.compass.ie/mapview-config-docs/tooltipsConfig.html#tooltipsconfig-json

Properties matched OL vector feature properties, and an optional alias was allowed to have a more "human-friendly" field title.

 ```
           "tooltipsConfig": [
                {
                    "property": "PropertyName1"
                }, 
                {
                    "alias": "Property Name 1", 
                    "property": "PropertyName1"
                }, 
```

PR #30, opened 2018-11-07
 Add 'singleclick' handler to MapPanel

chrismayer:
This adds a `singleclick` handler on the map which fires an event `cmv-mapclick` with all clicked vector features and their corresponding layers.

The event propagates the clicked features and their corresponding layers, like 

`[{feature: aFeat, layer: layerOfFeat}, ...]`

and the original OpenLayers event. 

In order to react on the map click in an application, as in #28, just use the following code skeleton:

```javascript
var mapPanel = CpsiMapview.view.main.Map.guess();
mapPanel.on('cmv-mapclick',  function(clickedObjs, evt) {
    // do sth. with the features here
    console.log(clickedObjs, evt);
});
```

### DIFF
--- a/app/view/main/Map.js
+++ b/app/view/main/Map.js
@@ -56,6 +56,18 @@ Ext.define('CpsiMapview.view.main.Map', {
      * @param {ol.MapBrowserEvent)} evt The original 'singleclick' event of OpenLayers
      */
 
+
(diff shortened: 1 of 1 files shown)


chrismayer:
> enhance this with pointermove changing of the cursor

This maybe can be done with #27

Issue #10, opened 2018-08-31
[LayerFactory] Support all config options of Layer-JSON


chrismayer:
At the moment not all config options of the MapView Configuration are supported in the LayerFactory. This should be done in the future.

geographika:
@chrismayer - the priority here would be **createWfs**

PR #33, opened 2018-11-08
Implement 'createWfs' function in LayerFactory (wo/ styling)

chrismayer:
This adds a function to create an OGC WFS layer from the corresponding JSON configuration. So far no custom styling of features is implemented.

See #10.

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -106,9 +106,75 @@ Ext.define('CpsiMapview.factory.Layer', {
         });
     },
 
+    /**
+     * Creates an OGC WFS layer
+     *
+     * @param  {Object} layerConf The configuration object for this layer
+     * @return {ol.layer.Vector}  WFS layer
+     */
     createWfs: function(layerConf) {
+        var url = layerConf.url;
 
-        Ext.log.info('Not implemented yet', layerConf);
+        var srid = 'EPSG:3857';
+        var mapPanel = CpsiMapview.view.main.Map.guess();
+        if (mapPanel) {
+            srid = mapPanel.olMap.getView().getProjection().getCode();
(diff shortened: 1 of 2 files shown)


app/factory/Layer.js
```diff
@@ -106,9 +106,77 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+     * Creates an OGC WFS layer
+     *
+     * @param  {Object} layerConf The configuration object for this layer
+     * @return {ol.layer.Vector}  WFS layer
+     */
     createWfs: function(layerConf) {
 
-        Ext.log.info('Not implemented yet', layerConf);
+        var url = layerConf.url;
+        if (url.indexOf() == -1) {
```
marcjansen:
```suggestion
        if (url.indexOf('?') == -1) {
```

app/factory/Layer.js
```diff
@@ -106,9 +106,77 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+            url+= '?';
```
marcjansen: nit-pick, space missing
```suggestion
            url += '?';
```

app/factory/Layer.js
```diff
@@ -106,9 +106,77 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+        } else if (!url.endsWith('?') && !url.endsWith('&')) {
+            url += '&';
+        }
```
marcjansen: Ther is a utility function in ExtJS that does this, which would probably be even better.

chrismayer:
Thanks for your catch and the hint with the Ext util function, @marcjansen. I reworked this by using `Ext.String.urlAppend`.

PR #34, opened 2018-11-08
Support ImageWMS and TileWMS in LayerFactory

### LINKED ISSUE #9: [LayerFactory] Check for support of ImageWMS and TileWMS
At the moment the layer-config type `wms` will create a layer with a source `ol.source.TileWMS`. We should check if there is a need to add layers with a source of type `ol.source.ImageWMS`.

chrismayer:
This adds support for single tile WMS layers (in addition to the tiled WMS layers) in the LayerFactory.

Closes #9

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -88,22 +88,46 @@ Ext.define('CpsiMapview.factory.Layer', {
      * @return {ol.layer.Tile}    WMS layer
      */
     createWms: function(layerConf) {
-        return new ol.layer.Tile({
-            name: layerConf.text,
-            source: new ol.source.TileWMS({
-                url: layerConf.url,
-                params: {
-                    'LAYERS': layerConf.serverOptions.layers,
-                    'TILED': true,
-                    'TRANSPARENT': true
-                },
-                crossOrigin: 'anonymous'
-            }),
-            visible: layerConf.openLayers.visibility,
-            minResolution: layerConf.openLayers.minResolution,
-            maxResolution: layerConf.openLayers.maxResolution,
-            opacity: layerConf.openLayers.opacity
-        });
+        var layer;
+        var singleTile = layerConf.openLayers.singleTile;
+
+        if (singleTile) {
+            layer = new ol.layer.Image({
+                name: layerConf.text,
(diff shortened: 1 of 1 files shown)


app/factory/Layer.js
```diff
@@ -88,22 +88,45 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+                        'TRANSPARENT': true
+                    },
+                    crossOrigin: 'anonymous'
+                }),
+                visible: layerConf.openLayers.visibility,
+                minResolution: layerConf.openLayers.minResolution,
+                maxResolution: layerConf.openLayers.maxResolution,
+                opacity: layerConf.openLayers.opacity
+            });
+        }
```
marcjansen: This has some duplicated code when basically everything is the same, except for the constructors that are used and one param (`'TILED': true`) in a config object. (I assume that `crossOrigin` should be present in both `if`-branches) 

This is perfectly fine as it is, in case you want to change it to have less duplicated parts, I agree. I'll let you decide which way you go, because in the current way the readability probably is a bit better.

app/factory/Layer.js
```diff
@@ -88,22 +88,45 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+            layer = new ol.layer.Tile({
+                name: layerConf.text,
+                source: new ol.source.TileWMS({
+                    url: layerConf.url,
+                    params: {
+                        'LAYERS': layerConf.serverOptions.layers,
+                        'TILED': true,
+                        'TRANSPARENT': true
+                    },
+                    crossOrigin: 'anonymous'
```
marcjansen: This should also be set for the `singleTile`-case.

app/factory/Layer.js
chrismayer: I agree and I plan to harmonize some parts of the property applying at all, since some are duplicated in other functions as well. For the moment it gives a better overview which config properties are already processed.

app/factory/Layer.js
chrismayer: Right, I added this.

chrismayer:
Thanks for the review @marcjansen! I will merge now.

PR #35, opened 2018-11-08
Add more layer types to LayerFactory

chrismayer:
This adds the following layer types to the `LayerFactory`:

  - Bing
  - World Wind (BlueMarble-200412)
  - ArcGIS REST

See #10

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -177,9 +177,30 @@ Ext.define('CpsiMapview.factory.Layer', {
         return wfsLayer;
     },
 
+    /**
+     * Creates a Bing layer
+     *
+     * @param  {Object} layerConf The configuration object for this layer
+     * @param  {String} type      The Bing layer type, e.g. 'Aerial'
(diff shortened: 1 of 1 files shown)


app/factory/Layer.js
```diff
@@ -226,21 +247,69 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+                    params: layerConf.serverOptions || {}
+                }),
+                visible: layerConf.openLayers.visibility,
+                minResolution: layerConf.openLayers.minResolution,
+                maxResolution: layerConf.openLayers.maxResolution,
+                opacity: layerConf.openLayers.opacity
+            });
+        }
+
+        return layer;
```
marcjansen: Same comment wrt singleTile and duplicated code. I'll let you decide.

chrismayer: See my comment in #34

chrismayer:
Thanks for the review @marcjansen! I will merge now.

PR #36, opened 2018-11-09
 Custom Tooltips for Vector Layers

### LINKED ISSUE #27: Add Custom Tooltips for Vector Layers

chrismayer:
This adds custom tooltips for hovered vector features on the map. The tooltips can be enabled if the corresponding layer has the property `tooltipsConfig` in its JSON-configuration:

grafik

Closes #27.

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -204,10 +204,6 @@ Ext.define('CpsiMapview.factory.Layer', {
 
             // show / hide on appropriate events
             mapPanel.on('cmv-map-pointerrest', function(hoveredObjs, evt) {
-                // hide if no feature was hit on pointerrest
-                if (hoveredObjs.length === 0) {
-                    toolTip.hide();
-                }
                 // show tooltip with feature attribute information
                 Ext.each(hoveredObjs, function (hoveredObj) {
                     if (hoveredObj.layer &&
@@ -218,8 +214,14 @@ Ext.define('CpsiMapview.factory.Layer', {
                 });
             });
 
-            mapPanel.on('cmv-map-pointerrestout', function () {
+            // hide tooltip if mouse moves again
+            mapPanel.on('cmv-map-pointermove', function () {
                 toolTip.hide();
+            });
+
+            // hide all tooltips if cursor leaves map
+            mapPanel.on('cmv-map-pointerrestout', function () {
+                CpsiMapview.view.layer.ToolTip.clear();
             });
         }
 
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -210,7 +210,9 @@ Ext.define('CpsiMapview.factory.Layer', {
                 }
                 // show tooltip with feature attribute information
                 Ext.each(hoveredObjs, function (hoveredObj) {
-                    if (hoveredObj.layer.toolTip) {
+                    if (hoveredObj.layer &&
+                          hoveredObj.layer.id === wfsLayer.id &&
+                          hoveredObj.layer.toolTip) {
                         hoveredObj.layer.toolTip.draw(hoveredObj.feature, evt);
                     }
                 });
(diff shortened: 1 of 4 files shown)


app/factory/Layer.js
```diff
@@ -190,12 +190,35 @@ Ext.define('CpsiMapview.factory.Layer', {
...
-            //     https://github.com/meggsimum/cpsi-mapview/issues/27
+            // create a custom toolitp for this layer
+            var toolTip = Ext.create('CpsiMapview.view.layer.ToolTip', {
+                toolTipConfig: layerConf.tooltipsConfig,
+                layer: wfsLayer
+            });
+            wfsLayer.toolTip = toolTip;
+
+            // show / hide on appropriate events
+            mapPanel.on('cmv-map-pointerrest', function(hoveredObjs, evt) {
```
marcjansen: Do we need to check if the hovered feat is from this layer?

chrismayer: This is implicitly done in Line 214. The tooltip, which is rendered, is bound to the layer and is only available if the layer is 'hoverable'.

marcjansen: What I meant is the following: if we duplicate the layer GAS WFS and call it GAS WFS 2, and put in a log in line 2017 which prints out the `hoveredObjs`, you'll see, that we get two logs (expected), but both of them have two features (rather unexpected, right?). A possible (untested) fix could be to check `hoveredObj.layer` against `wfsLayer`. Is this more clear now?

marcjansen:
Two other remarks:

**Popups of two or more layers overlapping**

image

This is what can happen when we hav two layers that are hoverable, not sure if this can be circumvented somehow. It can also of course be, that the smaller tooltipo lies below the bigger one, in which case you do not see it.

**Popups of different layers at the same time**

image

same here, not sure what dto do here and whether this is an actual issue.

Nothing blocking, just sth. we need to think about / document?

app/view/layer/ToolTip.js
```diff
@@ -0,0 +1,153 @@
...
+            html = me.layer.formatFunction(feature);
+        } else {
+            html = me.formatFunction(feature);
+        }
+
+        // set HTML content
+        me.setHtml(html);
+
+        // show tooltip near mouse pointer
+        var screenX = evt.originalEvent.x +  me.offsetX;
```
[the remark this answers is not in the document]
chrismayer: Done in 51cfef734b52.

app/factory/Layer.js
chrismayer: Done in 51cfef734b52.

geographika:
Looks very nice!

@chrismayer - just a note on the old logic - it was also based on the layer order (or technically z-index), so there would only be one tooltip at a single location. It is simple enough for users to turn off layers if they 
want to focus on a particular vector layer and its tooltips.

marcjansen:
Thanks for the feedback, Seth

chrismayer:
Thanks for your reviews @marcjansen and thanks for the feedback @geographika! 

So this is good to go. I'll merge now.

Issue #21, opened 2018-09-13
Add Measure Tools


geographika:
![image](https://user-images.githubusercontent.com/490840/45495329-5c274680-b773-11e8-98fb-12501981e077.png)

Reuse http://terrestris.github.io/BasiGX/master/docs/#!/api/BasiGX.view.button.Measure ?

Originals used OpenLayers.Control.Measure and both line and polygons were supported. 

https://github.com/compassinformatics/mapview/blob/master/src/gis/MeasureAreaToolAction.js
https://github.com/compassinformatics/mapview/blob/master/src/gis/MeasureAreaTool.js

https://github.com/compassinformatics/mapview/blob/master/src/gis/MeasureLineTool.js
https://github.com/compassinformatics/mapview/blob/master/src/gis/MeasureLineToolAction.js

marcjansen:
Yes, the mentioned button can be used

marcjansen:
Fixed in #37

PR #37, opened 2018-11-12
Measure tools

ahennr:
This PR adds BasiGX measure tools to to header toolbar of cpsi-mapview.

Currently the following (Font Awesome) icons are used for buttons:
image

### DIFF
--- a/app/controller/button/MeasureButton.js
+++ b/app/controller/button/MeasureButton.js
@@ -1,5 +1,5 @@
 /**
- * This class is the controller for the measure buttons for the application. *
+ * This class is the controller for the measure buttons for the application.
  */
 Ext.define('CpsiMapview.controller.button.MeasureButtonController', {
     extend: 'Ext.app.ViewController',
--- a/app/controller/button/MeasureButton.js
+++ b/app/controller/button/MeasureButton.js
@@ -1,0 +1,8 @@
+/**
+ * This class is the controller for the measure buttons for the application. *
+ */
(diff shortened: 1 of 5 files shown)


ahennr:
Closes #21

app/controller/button/MeasureButton.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * This class is the controller for the measure buttons for the application. *
```
marcjansen: Extra `*`

```suggestion
 * This class is the controller for the measure buttons for the application.
```

app/view/main/Map.js
```diff
@@ -19,6 +23,39 @@ Ext.define('CpsiMapview.view.main.Map', {
         items: [{
             xtype: 'basigx-button-zoomtoextent',
             extent: [-1210762, 6688545, -600489, 7490828]
+        }, {
+            xtype: 'basigx-button-measure',
+            measureType: 'line',
+            toggleGroup: 'measure-tools',
+            viewModel: 'cmw_btn_measure',
+            controller: 'cmv_btn_measure',
+            glyph: 'xf068@FontAwesome',
```
marcjansen: I see, these buttons need to be changed, this can be done in a follow up PR

app/controller/button/MeasureButton.js
ahennr: fixed in b0c880a39a

Issue #25, opened 2018-09-13
Add Coordinates to Map


geographika:
Add to toolbar at bottom of the map in EPSG:2157

![image](https://user-images.githubusercontent.com/490840/45495873-7281d200-b774-11e8-8a3c-24a16fb58551.png)

Can’t find an example at: https://openlayers.org/en/latest/examples/

See https://github.com/compassinformatics/mapview/blob/master/src/gis/CustomMap.js
**addMouseMoveCoord**

Latest mapview version has a full component similar to below:

![image](https://user-images.githubusercontent.com/490840/45495903-84fc0b80-b774-11e8-8d60-020d58a24af1.png)

See https://github.com/compassinformatics/mapview/blob/master/src/gis/Coordinates.js

chrismayer:
This OL control should do the trick: http://openlayers.org/en/v4.6.5/apidoc/ol.control.MousePosition.html
Here as live example: https://openlayers.org/en/v4.6.5/examples/mouse-position.html

geographika:
That looks perfect. 
I'm guessing to get the Irish projection EPSG:2157 that http://proj4js.org/ is needed. 
What is the best way of adding this dependency? Simply copying into the lib folder?

marcjansen:
For proj4 we might be able to use npm... otherwise a git repo as submodule. copying a built js file is the last resort... and not a bad choice by itself.

marcjansen:
Fixed in #39

PR #39, opened 2018-11-13
WIP: 25 coordinates to map

ahennr:
Depends on https://github.com/terrestris/BasiGX/pull/348


ahennr:
Closes #25

index.html
```diff
@@ -46,6 +46,9 @@
     </script>
     -->
 
+    <!-- Add Proj4js -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/proj4js/2.5.0/proj4.js"></script>
```
chrismayer: I'd prefer to include proj4.js in the `app.json` as it is done with OpenLayers. The `app.json` (or better its "js" section) also accepts remote files.

app/controller/MapController.js
```diff
@@ -38,6 +38,7 @@ Ext.define('CpsiMapview.controller.MapController', {
 
     /**
      * Initialize measure buttons
+     * @param {Ext.button.Button} btn The measure button
```
chrismayer: This has nothing to do with this PR, or?

app/controller/MapController.js
```diff
@@ -49,4 +50,5 @@ Ext.define('CpsiMapview.controller.MapController', {
             btn.getViewModel().get('polygonMeasureAreaTooltip');
         btn.tooltipStr = tTipStr;
     }
+
```
chrismayer: This has nothing to do with this PR, or?

app/controller/MapController.js
ahennr: No, it doesn't. However, should I create a certain PR for this "minor" change in documentation or should I revert this?

index.html
ahennr: Alright, moved dependency to `app.json` in 4db9297

app/controller/MapController.js
marcjansen: If easily done, then yes. Don't let this block you though.

app/controller/MapController.js
marcjansen: Same here.

Issue #18, opened 2018-09-13
Add Map Zoom in and Out Tools


geographika:
![image](https://user-images.githubusercontent.com/490840/45494933-73b1ff80-b772-11e8-93f1-55bd4a3207f6.png)

Zoom In Tool 
Zoom Out Tool

These were originally based in OL2 on OpenLayers.Control.ZoomBox()

See https://github.com/compassinformatics/mapview/blob/master/src/gis/NavigationTools.js

OL4 has https://openlayers.org/en/latest/examples/extent-interaction.html but couldn’t find an exact equivalent to OL2.

marcjansen:
These are available in BasiGX

marcjansen:
We have these now. If in the future we need e.g. zoo by rectangle we can either reopen or create a new issue.

PR #43, opened 2018-11-15
Include zoom buttons

marcjansen:
This includes a zoom in and out button, that both react on a single click and either zoom in or out.

See #18. 

@geographika Is this what was envisioned in #18?

We opted out of rectangle zoom-in and out AFAICT

zoom2


geographika:
I think it was decided to go with simple step zoom buttons for the prototype. 

I may come back to this in a future iteration as these seem to be the tools users like most to interact with the map - I guess it is the fact they centre and zoom in the same interaction.

marcjansen:
Thanks everyone for clarification and thumbs-ups

Issue #23, opened 2018-09-13
Add Map Permalinks


geographika:
GeoExt2 had the following class https://geoext.github.io/geoext2/docs/#!/api/GeoExt.state.PermalinkProvider
No equivalent in GeoExt3?

Basigx has http://terrestris.github.io/BasiGX/master/docs/#!/api/BasiGX.view.form.Permalink
Is there a live example of this?

Mapview had a simple HTML link that would open the map in a new tab:

![image](https://user-images.githubusercontent.com/490840/45495473-9d1f5b00-b773-11e8-98a1-33a162db3eda.png)

marcjansen:
There are several approaches here, OpenLayers has one nkce example with super simple code. We should talk about the actual requurements.

The BasiGX implementation has some dependency on an external state provider / backend AFAIK, but I need to check

chrismayer:
There is still a [StateProvider](https://docs.sencha.com/extjs/6.2.0/classic/Ext.state.Provider.html) base class in ExtJS 6 but only implementations for [Cookies](https://docs.sencha.com/extjs/6.2.0/classic/Ext.state.CookieProvider.html) and [LocalStorage](https://docs.sencha.com/extjs/6.2.0/classic/Ext.state.LocalStorageProvider.html) so far.

geographika:
Previous functionality was based around  GeoExt.state.PermalinkProvider and OpenLayers.Control.Permalink. 

The only state recorded was x,y, and zoom level - e.g. foo?map_x=5&map_y=45&map_zoom=10
Full application state would be very tricky to implement so the above would be the priority. Layer visibility would be a "nice to have".

PR #44, opened 2018-11-15
Add map permalink functionality

chrismayer:
This records the state for map's zoom, center and rotation and adds it as hash to the current URL.

Inspired by http://openlayers.org/en/v4.6.5/examples/permalink.html

See #23.

### DIFF
--- a/app/view/main/Map.js
+++ b/app/view/main/Map.js
@@ -109,6 +109,13 @@ Ext.define('CpsiMapview.view.main.Map', {
     shouldUpdatePermalink: true,
 
     /**
+     * Flag to steer if center coordinates in the permalink should be rounded or
+     * not
+     * @config {Boolean}
+     */
+    roundPermalinkCoords: true,
+
+    /**
      * @event cmv-mapclick
      * Fires when the OL map is clicked.
      * @param {CpsiMapview.view.main.Map} this
@@ -302,11 +309,24 @@ Ext.define('CpsiMapview.view.main.Map', {
             me.shouldUpdatePermalink = true;
             return;
         }
+
         var mapState = me.getState();
+
+        // check if we have to round the coords (if no coordinates in deegrees)
+        var doRound =
+            me.roundPermalinkCoords &&
+            me.olMap.getView().getProjection().getUnits() !== 'degrees';
+        var centerX = mapState.center[0];
+        var centerY = mapState.center[1];
+        if (doRound) {
+            centerX = Math.round(centerX * 100) / 100;
+            centerY = Math.round(centerY * 100) / 100;
+        }
+
         var hash = '#map=' +
             mapState.zoom + '/' +
-            Math.round(mapState.center[0] * 100) / 100 + '/' +
-            Math.round(mapState.center[1] * 100) / 100 + '/' +
+            centerX + '/' +
+            centerY + '/' +
             mapState.rotation;
 
         // push the state into the window history (to navigate with browser's
--- a/app/view/main/Map.js
+++ b/app/view/main/Map.js
@@ -93,6 +93,21 @@ Ext.define('CpsiMapview.view.main.Map', {
      */
     addScaleBarToMap: true,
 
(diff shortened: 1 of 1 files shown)


app/view/main/Map.js
```diff
@@ -93,6 +93,21 @@ Ext.define('CpsiMapview.view.main.Map', {
      */
     addScaleBarToMap: true,
 
+    /*
+     * Flag that enables/disables permalink functionality
+     * @config {Boolean}
```
ahennr: This should be `@property` as well, shouldn't it?

app/view/main/Map.js
```diff
@@ -177,6 +192,119 @@ Ext.define('CpsiMapview.view.main.Map', {
...
+
+    /**
+     * Applies the map state (center, zoom, rotation) to this map.
+     *
+     * @param  {Object} mapState The map state object
+     */
+    applyState: function (mapState) {
+        var me = this;
+
+        me.olMap.getView().setCenter(mapState.center);
```
ahennr: `mapState` could be an empty object / null here if the permalink could not be read, or am I wrong? Could you please add a null check?

app/view/main/Map.js
```diff
@@ -177,6 +192,119 @@ Ext.define('CpsiMapview.view.main.Map', {
...
+        if (!me.shouldUpdatePermalink) {
+            // do not update the URL when the view was changed in the 'popstate'
+            // handler
+            me.shouldUpdatePermalink = true;
+            return;
+        }
+        var mapState = me.getState();
+        var hash = '#map=' +
+            mapState.zoom + '/' +
+            Math.round(mapState.center[0] * 100) / 100 + '/' +
```
ahennr: Will map projection always be `EPSG:3857`? If so, resricting to 2 digits is fine IMHO. But if it is change to `EPSG:4326`,  for example (unit degree) this would lead to a loss of information of center position.

app/view/main/Map.js
chrismayer: No, this is a config which can be passed from outside for initialization. A change of this option after the initialization would have no effect.

app/view/main/Map.js
chrismayer: Yep, added check in 5aac0ec

app/view/main/Map.js
chrismayer: I made the rounding configurable, see 322b37d

app/view/main/Map.js
```diff
@@ -177,6 +199,138 @@ Ext.define('CpsiMapview.view.main.Map', {
...
+     */
+    registerPermalinkEvents: function () {
+        var me = this;
+
+        // update permalink when visible map state changes
+        me.olMap.on('moveend', me.updatePermalink, me);
+
+        // restore the view state when navigating through the history, see
+        // https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate
+        window.addEventListener('popstate', function(event) {
```
marcjansen: According to the link above this is supported for IE10 and higher, and also in every sane browser. This [bugreport](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/3740423/) mentions issues in IE 11 / Edge …

I think it would be good if we culd test this in an IE, and also if @geographika may confirm this works for him. I want this to be in, so he can easily test this.

geographika: @marcjansen - tested in Edge and IE11, both work fine. 
I tried IE10 with the emulator option in IE11 but the page doesn't load (with or without permalinks). 
For the prototype it is fine to target IE11 and Edge only. So LGTM.

marcjansen: Thanks a ton for this info, @geographika.

chrismayer: Great, thanks for testing @geographika!

chrismayer:
Thanks for your reviews @ahennr and @marcjansen! I'll merge now.

PR #49, opened 2018-11-16
Time range slider

ahennr:
This PR enhances previously introduces time-slider by range selection. If range filter is noch active (corresponding checkbox not checked), the second thumb is disabled and hidden by CSS.

In addition, a util is introduced which can be used to get all "time-dependent" layers from the map.

In order to test slider with a running example, please use docker-compose config introduced in 24ec7ae (needed to omit CORS problems...). In `nginx.conf`, fill `proxy_pass` configs with appropriate values.


app/controller/panel/TimeSlider.js
```diff
@@ -6,59 +6,101 @@ Ext.define('CpsiMapview.controller.panel.TimeSlider', {
...
 
         slider.setMaxValue(timeRangeObj.maxValue);
         slider.setMinValue(timeRangeObj.minValue);
+        var values = [timeRangeObj.rangeLower, timeRangeObj.rangeUpper];
+        slider.setValue(values);
 
-        viewModel.setData({
-            timeRangeObj: timeRangeObj
-        });
+        // hide thumb if range is noit checked
```
marcjansen: typo

```suggestion
        // hide thumb if range is not checked
```

app/controller/panel/TimeSlider.js
```diff
@@ -6,59 +6,101 @@ Ext.define('CpsiMapview.controller.panel.TimeSlider', {
...
+    },
+
+    /**
+     * Function to set the time to time dependent layers after slider initialization
+     * @param {Ext.slider.Multi} slider The time slider component
+     */
+    setTimeOnLayers: function (slider) {
+        var me = this;
+        var isRange = me.getViewModel().get('isRange');
+        this.onTimeChanged(slider, isRange);
```
marcjansen:
```suggestion
        me.onTimeChanged(slider, isRange);
```

app/controller/panel/TimeSlider.js
```diff
@@ -6,59 +6,101 @@ Ext.define('CpsiMapview.controller.panel.TimeSlider', {
...
+     * @param {Boolean} isRange Is range query or not
+     *
+     * @returns {String} The formatted time string
+     */
+    formatWmsDateString: function (values, isRange) {
+        var me = this;
+        // we must check for min / max here since the thumbs are not contrained
+        var startDateStr = me.getDateStringForSliderValue(Ext.Array.min(values));
+        var endDateStr = me.getDateStringForSliderValue(Ext.Array.max(values));
+        // TODO: format can be layer dependent....
```
marcjansen: We can read this out from config and store in the layer. See the line 74 and those coming after that in the Factory for layers.

app/util/Layer.js
```diff
@@ -0,0 +1,39 @@
...
+    ],
+    statics: {
+        TIME_DEPENDENT_ATTRIBUTE_NAME: 'isTimeDedendent',
+
+        /**
+         * Function to get all time dependent layers in map
+         * @param {ol.collection} collection (optional) an OpenLayers layer collection
+         * @returns {ol.Layer[]} An array of time dependent layers
+         */
+        getTimeDependentLayers: function(collection) {
```
marcjansen: It might make sense to add method to the layer util of BasiGX `getLayersBy()` (we already have `getLayerBy`), can you change this probably?

ahennr: Moved to BasiGX, see here https://github.com/terrestris/BasiGX/pull/354

app/controller/panel/TimeSlider.js
ahennr: Alright, I'll create an issue for that

PR #51, opened 2018-11-19
Add Back and Forward Buttons

### LINKED ISSUE #20: Add Back and Next Buttons
image

Originals based on OpenLayers.Control.NavigationHistory()
See https://github.com/compassinformatics/mapview/blob/master/src/gis/NavigationTools.js

chrismayer:
This adds two [HistoryButtons](https://github.com/terrestris/BasiGX/blob/master/src/view/button/History.js) of BasiGX to navigate through the browser history. This allows to restore previous map stages.

The needed permalinks are managed by a [GeoExt.state.PermalinkProvider](https://github.com/geoext/geoext3/blob/master/src/state/PermalinkProvider.js).

Based on https://github.com/terrestris/BasiGX/pull/353 and https://github.com/geoext/geoext3/pull/451

Closes #20.

### DIFF
--- a/app/view/main/Map.js
+++ b/app/view/main/Map.js
@@ -7,6 +7,7 @@ Ext.define('CpsiMapview.view.main.Map', {
 
     requires: [
         'GeoExt.component.Map',
+        'GeoExt.state.PermalinkProvider',
 
         'CpsiMapview.model.button.MeasureButton',
         'CpsiMapview.controller.button.MeasureButtonController',
@@ -64,6 +65,8 @@ Ext.define('CpsiMapview.view.main.Map', {
         xtype: 'gx_map',
         pointerRest: true,
         pointerRestInterval: 500,
+        stateful: true,
+        stateId: 'cmv_mapstate',
         map: new ol.Map({
             // layers will be created from config in initComponent
             layers: [],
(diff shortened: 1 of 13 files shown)


app/view/main/Map.js
```diff
@@ -54,6 +56,14 @@ Ext.define('CpsiMapview.view.main.Map', {
             listeners: {
                 afterrender: 'initializeMeasureBtn'
             }
+        }, {
+            xtype: 'basigx-button-history',
+            direction: 'BACK',
```
ahennr: Case-sensitive?

app/view/main/Map.js
```diff
@@ -54,6 +56,14 @@ Ext.define('CpsiMapview.view.main.Map', {
...
+            glyph: 'xf104@FontAwesome'
+        }, {
+            xtype: 'basigx-button-history',
+            direction: 'FORWARD',
```
ahennr: See above comment

app/view/main/Map.js
chrismayer: Yep, is more or less treated like a constant in the BasiGX class. Maybe I'll introduce a static member later, which can be used for this (instead of typing 'BACK' / 'FORWARD').

chrismayer:
Thanks for the review @ahennr! I'll merge once travis is happy.

PR #54, opened 2018-11-19
Add a Map Toolbar class

chrismayer:
This moves the map toolbar to an own class `CpsiMapview.view.toolbar.MapTools`. This makes it easier to maintain and it harmonizes the code structure with the bottom bar.

The controller logic remained in the MapController class, we might want to think about to move it over to new a new Controller.

### DIFF
--- a/app/view/main/Map.js
+++ b/app/view/main/Map.js
@@ -21,7 +21,8 @@ Ext.define('CpsiMapview.view.main.Map', {
     controller: 'cmv_map',
 
     dockedItems: [{
-        xtype: 'cmv_maptools'
+        xtype: 'cmv_maptools',
+        dock: 'top',
     }, {
         xtype: 'cmv_mapfooter',
         dock: 'bottom'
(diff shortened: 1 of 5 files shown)


app/view/main/Map.js
```diff
@@ -1,5 +1,5 @@
 /**
- * This view is an example list of people.
+ * Main map view.
```
marcjansen: Ahemm, good catch :wink:

app/view/toolbar/MapTools.js
```diff
@@ -0,0 +1,65 @@
...
+            extent: [-1210762, 6688545, -600489, 7490828]
+        }, {
+            xtype: 'basigx-button-zoomin'
+        }, {
+            xtype: 'basigx-button-zoomout'
+        }, {
+            xtype: 'basigx-button-measure',
+            measureType: 'line',
+            toggleGroup: 'measure-tools',
+            viewModel: 'cmw_btn_measure',
```
marcjansen: BTW, is this the right viewmpdel alias?! :thinking:

PR #64, opened 2018-11-26
SLD support for WFS layers

### LINKED ISSUE #57: Apply Styling to Vector Layers
In the original mapview a WFS layer config could include a "sld" parameter which loaded an SLD file from a "sldUrl" property. This allowed styling to be applied from an SLD file. SLDs could be created dynamically, but it was faster to generate them on deployment and load these cached files). 

See https://github.com/compassinformatics/mapview/blob/master/src/gis/SldMixin.js and createWFS in LayerFactory.js

Does the https://github.com/terrestris/BasiGX/blob/master/src/util/SLD.js class allow an SLD file to be applied to a ol4 vector layer? If so then applying this in the Layer Factory may be a possible option. 

A more complex "slds" option built a radio button interface to allow switching between styling on the same vector layer to avoid reloading. This is likely to be a future requirement.

chrismayer:
This adds SLD-based styling functionality for WFS layers. Parsing of the SLD and transformation to OL Styles is done by [geostyler-sld-parser](https://github.com/terrestris/geostyler-sld-parser) and [geostyler-openlayers-parser](https://github.com/terrestris/geostyler-openlayers-parser) of the [GeoStyler](https://github.com/terrestris/geostyler) ecosystem.

image

Closes #57


app/factory/Layer.js
```diff
@@ -523,6 +522,40 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+        Ext.Ajax.request({
+            url: sldUrl,
+            method: 'GET',
+            success: function(response) {
+                var sldXml = response.responseText;
+                var sldParser = new GeoStylerSLDParser.SldStyleParser();
+                var olParser = new GeoStylerOpenlayersParser.OlStyleParser(ol);
+
+                sldParser.readStyle(sldXml)
+                    .then(function (gs) {
```
marcjansen: Do we need a Promise polyfill?

marcjansen: In case this doesn't work as expected, we might e.g. sinply use this [polyfill](https://github.com/getify/native-promise-only/blob/master/README.md) or one of the alternatives.

chrismayer: Yes, we should be aware that there might be a need for polyfills, which should be included in the `app.json`, e.g. for Promise and History (see #51) .

chrismayer:
Thanks for the reviews @marcjansen and @ahennr and thank you for the hint with the build-script. We surely can include this in a later stage.

PR #67, opened 2018-11-29
Update BasiGX to current master

marcjansen:
Title says it all. This also removes the annoying warnings when building the application.

Please review.


marcjansen:
Thanks for the review,

PR #83, opened 2019-04-17
Add "Refresh Layer" tool as item in layer context menu

### LINKED ISSUE #77: Add a Refresh Layer Plugin
Port functionality from https://github.com/compassinformatics/mapview/blob/master/src/gis/tree/contextmenu/LayerRefresh.js and add as layer plugin (see #76)

chrismayer:
This adds a "Refresh Layer" tool as `MenuItem` in the `LayerTree`'s ContextMenu.

Closes #77

### DIFF
--- a/app/view/menuitem/LayerRefresh.js
+++ b/app/view/menuitem/LayerRefresh.js
@@ -1,0 +1,64 @@
+/**
+ * MenuItem to force redraw of a layer.
+ *
+ * @class CpsiMapview.view.menuitem.LayerRefresh
+ */
+Ext.define('CpsiMapview.view.menuitem.LayerRefresh', {
+    extend: 'Ext.menu.Item',
+    xtype: 'cmv_menuitem_layerrefresh',
+    requires: [],
+
+    /**
+     * The connected layer for this item.
+     *
+     * @cfg {ol.layer.Base}
+     */
+    layer: null,
+
+    /**
+     * Text shown in this MenuItem
+     * @cfg {String}
+     */
+    text: 'Refresh',
+
+
+    /**
+     * @private
+     */
+    initComponent: function () {
+        var me = this;
+        var allowRefresh = false;
+
+        if (me.layer) {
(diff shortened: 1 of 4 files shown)


app/view/menuitem/LayerRefresh.js
```diff
@@ -0,0 +1,64 @@
...
+        if (source.updateParams) {
+            var params = source.getParams();
+            params.noCache = new Date().getMilliseconds();
+            source.updateParams(params);
+            source.refresh();
+        }
+
+        // mostly WFS layers
+        if (source instanceof ol.source.Vector) {
+            source.clear();
```
weskamm: shouldn't this be also a call of `refresh` instead of clear? Else the features will be gone

chrismayer: Calling clear() forces the reload of the underlying data source. A refresh() does not query the source again. But you are right in case of having a vector layer with non-remote features they will be gone. Maybe we restrict this to WFS layers (however we detect them)?

chrismayer:
Thanks for the review @weskamm!

I added some logic to detect the WFS layers and the `clear` is only called for them. For other layers / sources only `refresh` is called.

Would you please have another look? TIA!

app/view/menuitem/LayerRefresh.js
weskamm: hmm ok, did not have that in mind, then clear should be ok.
For wfs detextion: i think every non-local vector layer should have an url or urlfunction defined, which is some kind of identifier for WFS...

chrismayer: > For wfs detextion: i think every non-local vector layer should have an url or urlfunction (or loader) defined, which is some kind of identifier for WFS...

Unfortunately not. You can implement a custom `loader`. The URL is then undefined and a `loader` has no getter.  That's why I decided to introduce this custom property on the WFS layer.

chrismayer:
Thanks again for the review. Will merge now.

PR #84, opened 2019-04-18
Add "Labels Layer" tool as item in layer context menu 

### LINKED ISSUE #78: Add a Label Layer Plugin
See https://github.com/compassinformatics/mapview/blob/master/src/gis/tree/contextmenu/LayerLabels.js for original design. 

It adds a checkbox to the layer context menu, which adds in an additional STYLE to the layer parameter allowing labels to be displayed. A more generic approach allowing multiple styles to be added/removed on the layer may be preferable.  

See #76 re plugins.

chrismayer:
This adds a "Labels Layer" tool as `MenuItem` in the LayerTree's `ContextMenu`.

Currently contains the changes of #83 since this is based on that and is not merged yet. Please ignore the changes of #83. Main class to review is  `app/view/menuitem/LayerLabels.js`

Closes #78

### DIFF
--- a/app/view/menuitem/LayerLabels.js
+++ b/app/view/menuitem/LayerLabels.js
@@ -1,0 +1,122 @@
+/**
+ * MenuItem showing a checkbox in the layer context menu that allows
+ * a separate label layer to be displayed alongside the feature layer.
+ *
+ * @class CpsiMapview.view.menuitem.LayerLabels
+ */
+Ext.define('CpsiMapview.view.menuitem.LayerLabels', {
+    extend: 'Ext.menu.CheckItem',
+    xtype: 'cmv_menuitem_layerlabels',
+    requires: [],
+
+    /**
+     * The connected layer for this item.
+     * @cfg {ol.layer.Base}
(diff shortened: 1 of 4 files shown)


app/view/menuitem/LayerLabels.js
```diff
@@ -0,0 +1,122 @@
+/**
+ * MenuItem showing a checkbox in the layer context menu that allows
+ * a separate label layer to be displayed alongside the feature layer.
+ *
+ * @class CpsiMapview.view.menuitem.LayerLabels
+ */
+Ext.define('CpsiMapview.view.menuitem.LayerLabels', {
+    extend: 'Ext.menu.CheckItem',
+    xtype: 'cmv_menuitem_layerlabels',
+    requires: [],
```
weskamm: can be removed

chrismayer:
Thanks for the review @weskamm! Will merge now.

> Maybe sometime we can add these classes as examples for geoext3, but this can happen after this PR...

Sure let's think about this later on.

Issue #74, opened 2019-03-08
Port Network Solver Tool


geographika:
Port the OL2 based tool at https://github.com/compassinformatics/mapview/blob/master/src/gis/Solver.js

This depends on: 
- https://github.com/compassinformatics/mapview/blob/master/src/gis/ServiceToolMixin.js

This tool relies on a backend service. Test JSON responses will be specified and example test files added to this ticket. A demo of the tool in action can be provided.

geographika:
@weskamm - excellent, this works really well - thanks. 
One piece of functionality that is missing is the ability to click on a point in the path to remove a node. This allows users to reroute with the tool without having to start from scratch again. 
Example screencast from the old system: https://drive.google.com/file/d/12gVY-cUvKFZ9J568Vr1nQWsq7__qfqpw/view

Is this something that can be added to the tool?

geographika:
Quick question on using more specialised versions of these tools - I presume the standard approach of sub-classing the DigitizeButton controller and view is the recommended approach, and override onResponseFeatures?

geographika:
A couple of final comments:

1. Can the point be removed by default if it does not find a road?

![image](https://user-images.githubusercontent.com/490840/56738736-7f848c00-676d-11e9-92d1-745f03a63502.png)

2. If the Ajax request is cancelled or timesout (rather than simply the request failing), then the "Loading" gif is never removed. Is this something that can be handled in Ext.Ajax.request ?

I could recreate this by restarting the web server and creating a large polygon as in the screencast below:

https://drive.google.com/file/d/13o3dLAYDrBZsWjVtPJ-IETwLtyPs_QJ1/view




https://drive.google.com/file/d/13o3dLAYDrBZsWjVtPJ-IETwLtyPs_QJ1/view

annarieger:
Both issues (not found road points and eternal loading mask) are fixed in #86. Please take a look @geographika

geographika:
@annarieger - thanks for these fixes - confirmed as fixed locally.

PR #86, opened 2019-04-29
Removes the last drawn feature on error and fixes a NPE

dnlkoch:
This removes the last drawn feature on an erroneous request and fixes access of an possibily unavailable key.

See #74.

Please review @annarieger.


app/controller/button/DigitizeButtonController.js
```diff
@@ -305,7 +309,15 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
...
+                me.removeLastDigitizeFeature();
+
+                var errorMessage = 'Error while requesting the API endpoint';
+
+                if (response.responseText && response.responseText.message) {
+                    errorMessage += ': ' + response.responseText.message;
+                }
+
+                BasiGX.error(errorMessage);
             }
```
marcjansen: Having this also in a standalone function would be even better, but this totally non-blocking.

PR #89, opened 2019-05-06
Add gazetteer components

chrismayer:
This PR adds two gazetteer components:

  - a Nominatim based geocoding combobox from GeoExt
  - a custom CPSI gazetteer / geocoding combobox (extending the `GeocoderCombo` of GeoExt)

Closes #70

PR #90, opened 2019-05-07
Add missing Gazetteer combo class

chrismayer:
This adds the missing Gazetteer combo class (needed for the changes introduced with #89).


chrismayer:
Thanks @weskamm!

PR #91, opened 2019-05-08
Do not show geocoding layers in LayerTree

chrismayer:
Avoids showing the result layers of the geocoding / gazetteer comboboxes in the LayerTree.


chrismayer:
Thanks for your review @annarieger!

Issue #71, opened 2019-03-08
Port StreetView Tool


geographika:
The current mapview project includes a custom OpenLayers2 tool that allows a user to click a point on a map which opens up a window with the Google StreetView for that location. A user can then move around the StreetView window and a corresponding marker icon on the main map moves in synch. 

Current code: https://github.com/compassinformatics/mapview/blob/master/src/gis/StreetViewTool.js
and https://github.com/compassinformatics/mapview/blob/master/src/gis/StreetViewToolAction.js

There may be issues with integrating Google with the new version of OpenLayers. 
The StreetView tool requires the blue lines displaying where StreetView is available rather than the Google base maps. It is currently unclear if these can be used directly without full Google/OpenLayers integration. 

The following library may be useful as a dependency: https://github.com/mapgears/ol3-google-maps

TO SEND - google API key

geographika:
The current implementation works vey well, and was easy to inegrate into another project. 
One minor display issue was that the SV controls are hidden by the ExtJS window title as shown below:

![image](https://user-images.githubusercontent.com/490840/58412518-e8199e00-8076-11e9-9f3e-bdb270995902.png)

chrismayer:
Correct, I could reproduce the layout issue and I opened a PR (#96) to fix this.

geographika:
Tested and integrated - thanks @chrismayer !

PR #92, opened 2019-05-08
Add StreetView Tool (wo/ Google layers) 

chrismayer:
This ports the StreetView tool (see #71), showing the SV panorama for the clicked position on the map. 

This does not show any Google layers on the map so far.


app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+     * Handles the 'toggle' event of the button (view).
+     *
+     * @param  {Ext.button.Button} btn The toggled button
+     * @param  {Boolean}           pressed New pressed state
+     * @private
+     */
+    onToggle: function (btn, pressed) {
+        var me = this;
+
+        // detect the overlay layer group (to add psoition layer)
```
weskamm: typo

app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+        // GMaps API missing at all
+        if (!Ext.isObject(window.google)) {
+            reactOnMissingGmapsApi();
+            return;
+        }
+        // GMaps API not usable due to missing API key or similar
+        window.gm_authFailure = function() {
+            reactOnMissingGmapsApi();
+            return;
+        };
```
marcjansen: Can this be simplified to 

```javascript
window.gm_authFailure = reactOnMissingGmapsApi;
```

?

app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+                displayInLayerSwitcher: false,
+                source: new ol.source.Vector(),
+                style: new ol.style.Style({
+                    image: new ol.style.Icon(({
+                        anchor: [0.5, 46],
+                        anchorXUnits: 'fraction',
+                        anchorYUnits: 'pixels',
+                        src: view.vectorIcon
+                    }))
+                })
```
marcjansen: Would be nice if only the style was configurable.

app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+                    }
+                }
+            });
+        }
+
+        // load the panorama for the given position
+        me.streetViewService.getPanoramaByLocation(latLng, 50, function (data, status) {
+            if (status === google.maps.StreetViewStatus.OK) {
+                var win = me.streetViewWin;
+                var title = view.svWinTitlePrefix + ' Image Date: ' + data.imageDate;
```
marcjansen: Probably best to make the text configurable? Probably also for the way you detect the Overlay group by name, above somewhere.

app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * Handles 'pov_changed' event of the SV panorama.
+     * Triggers #updatePositionFeature.
+     *
+     * @private
+     */
+    handlePovChanged: function () {
+        var me = this;
+        me.updatePositionFeature();
```
marcjansen: Aliasing unneeded here IMO. Also below.

app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+     *
+     * @return {Number} Heading in radiant
+     */
+    getHeadingRad: function () {
+        var me = this;
+        var heading = me.svPanorama.getPov().heading;
+
+        // normalize and convert to radiant
+        heading = me.normaliseDegrees(heading);
+        heading = heading / 180 * Math.PI;
```
marcjansen: I think there is a BasiGX util doing this... unsure though.

app/controller/button/StreetViewTool.js
```diff
@@ -0,0 +1,380 @@
...
+
+                // initially draw the position on the map and set heading
+                me.drawPositionFeature(me.gmapsLatLng2olCoord(latLng));
+                me.updatePositionFeature();
+
+                google.maps.event.addListener(
+                    me.svPanorama, 'pov_changed', me.handlePovChanged.bind(me));
+                google.maps.event.addListener(
+                    me.svPanorama, 'position_changed', me.handlePositionChanged.bind(me));
+
```
marcjansen: Will we need to unregister these if the relevant components get destroyed?

app/controller/button/StreetViewTool.js
marcjansen: I was thinking of theses meathods here: http://terrestris.github.io/BasiGX/master/docs/#!/api/BasiGX.view.button.Measure, but they do not help you exactly

app/controller/button/StreetViewTool.js
chrismayer: I suggest to leave this here beacuse of backward compatibility to the old version and more important
because it gives us the envelope for further things to be done on the certain events in the future.

app/view/button/StreetViewTool.js
```diff
@@ -0,0 +1,96 @@
...
+ * Button to open the Street View tool.
+ *
+ * @class CpsiMapview.view.button.StreetViewTool
+ */
+Ext.define('CpsiMapview.view.button.StreetViewTool', {
+    extend: 'Ext.button.Button',
+    xtype: 'cmv_streetview_tool',
+
+    requires: [
+        'CpsiMapview.controller.button.StreetViewToolController'
```
geographika: Should this be CpsiMapview.controller.button.StreetViewTool (based on the filename of the controller)?

chrismayer: Correct, thanks.

chrismayer:
Thanks everybody for your reviews. I addressed your comments (if not commented otherwise) and I additionally introduced an alert if no SV panorama is available at the clicked position.

PR #94, opened 2019-05-12
Add a numeric attribute slider

### LINKED ISSUE #97: New slider class with range options for specifying a numerical filter
A new slider class with range options – very similar to the TimeSlider class
https://github.com/meggsimum/cpsi-mapview/blob/master/app/controller/panel/TimeSlider.js
https://github.com/meggsimum/cpsi-mapview/blob/master/app/view/panel/TimeSlider.js

The component would allow multiple layers to be registered with it (allowing them to all be filtered by a range)

A property would be set on the slider to use as the filter field (this would be a numeric field)
When the slider is active (using a checkbox similar to the “Time range” above) a filter is applied to the registered layers

The filter would apply to WMS and WFS layers using the new OGCFilter class https://github.com/geoext/geoext3/blob/master/src/util/OGCFilter.js

marcjansen:
This adds a new slider that controls a numerical filter which will be applied to associated layers.

Originally wanted:
>    A new slider class with range options – very similar to the TimeSlider class
>        https://github.com/meggsimum/cpsi-mapview/blob/master/app/controller/panel/TimeSlider.js
>        https://github.com/meggsimum/cpsi-mapview/blob/master/app/view/panel/TimeSlider.js
>    The component would allow multiple layers to be registered with it (allowing them to all be filtered by a range)
>    A property would be set on the slider to use as the filter field (this would be a numeric field)
>    When the slider is active (using a checkbox similar to the “Time range” above) a filter is applied to the registered layers
>    The filter would apply to WMS and WFS layers using the new OGCFilter class https://github.com/geoext/geoext3/blob/master/src/util/OGCFilter.js

Implnted as such:

image

* slider can be enabled and disabled by clicking a checkbox next to it
* the slider has a config `numericField` for the actual filter field (I'd personally think this is better suited at the layer level, but this is how it was specified)
* a new layer config `numericitem` controls whether the layer is affected
* the slider has two freely movable thumbs to create a range which the becomes the filter
* the boundaries, increment and the current values can be configured with following configuration options:
  * `minValue`, default `0`
  * `maxValue`, default `15`
  * `currLowerValue`, default `5`
  * `currUpperValue`, default `11`
  * `increment`, default `0.5`
* boundaries are included (as it is the case with a BETWEEN filter)
* labelös can be configured with `sliderLabel` and `checkboxLabel`
* the initial state can be configured with the config `enabled` (default is `true`)
* Tested with WFS (see example app, layer accidents)
* Untested (due to data) with WMS
* also the linting rules are harder
* this also updates GeoExt and BasiGX as we need two of their recent chagnes herein (https://github.com/terrestris/BasiGX/pull/456 and https://github.com/geoext/geoext3/pull/519)

Here is a combined Filter with nuerical attribute filter, bbox and a date filter:

```xml
<ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
  <ogc:And>
    <ogc:PropertyIsGreaterThanOrEqualTo>
      <ogc:PropertyName>DATE</ogc:PropertyName>
      <ogc:Literal>2014-01-01</ogc:Literal>
    </ogc:PropertyIsGreaterThanOrEqualTo>
    <ogc:PropertyIsLessThanOrEqualTo>
      <ogc:PropertyName>DATE</ogc:PropertyName>
      <ogc:Literal>2014-12-31</ogc:Literal>
    </ogc:PropertyIsLessThanOrEqualTo>
    <And>
      <PropertyIsGreaterThanOrEqualTo>
        <ValueReference>Speed_Limit</ValueReference>
        <Literal>50</Literal>
      </PropertyIsGreaterThanOrEqualTo>
      <PropertyIsLessThanOrEqualTo>
        <ValueReference>Speed_Limit</ValueReference>
        <Literal>100</Literal>
      </PropertyIsLessThanOrEqualTo>
    </And>
    <ogc:BBOX>
      <ogc:PropertyName>msGeometry</ogc:PropertyName>
      <gml:Envelope xmlns:gml="http://www.opengis.net/gml" srsName="EPSG:3857">
        <gml:lowerCorner>-1181716.457288825 6920761.4149964275</gml:lowerCorner>
        <gml:upperCorner>-625254.8913727418 7117968.947972183</gml:upperCorner>
      </gml:Envelope>
    </ogc:BBOX>
  </ogc:And>
</ogc:Filter>
```

Please review, @geographika  @chrismayer @weskamm @annarieger 

Fixes #97

### DIFF
--- a/app/controller/panel/NumericAttributeSlider.js
+++ b/app/controller/panel/NumericAttributeSlider.js
@@ -1,0 +1,136 @@
+Ext.define('CpsiMapview.controller.panel.NumericAttributeSlider', {
+    extend: 'Ext.app.ViewController',
+
+    alias: 'controller.cmv_numericattributeslider',
+
+    /**
+     * Initially filter if enabled.
+     */
+    initSlider: function() {
+        if (this.getView().getEnabled()) {
+            this.applySliderEffects();
+        }
+    },
+
+    /**
+     * Returns those layers that the slider will affect.
+     */
+    getAffectedLayers: function() {
+        return BasiGX.util.Layer.getLayersBy('isNumericDependent', true);
+    },
+
+    /**
+     * Orders the values of the slider so that the first index always is the
+     * lower value, and the secind is at least equally big.
+     * @param {number[]} values
+     */
+    orderValues: function(values) {
+        var lower = Math.min(values[0], values[1]);
+        var upper = Math.max(values[0], values[1]);
+        return [lower, upper];
+    },
+
+    /**
+     * Returns the current range of the slider as tooltip.
+     * @param {Ext.slider.Thumb} thumb
+     */
+    getTipText: function(thumb) {
+        var slider = thumb.slider;
+        var draggedValue = thumb.value;
+        var thumbIdx = thumb.index;
+        var otherThumb = slider.thumbs[thumbIdx == 0 ? 1 : 0];
+        var lowerThumb = otherThumb.value <= draggedValue ? otherThumb : thumb;
+        var upperThumb = lowerThumb === thumb ? otherThumb : thumb;
+        var lower = lowerThumb.value;
+        var upper = upperThumb.value;
+        if (lower === draggedValue) {
+            lower = '<strong>' + lower + '</strong>';
+        }
+        if (upper === draggedValue) {
+            upper = '<strong>' + upper + '</strong>';
+        }
+        return lower + ' - ' + upper;
+    },
+
+    /**
+     * Handles the changing of the checked state of the active-checkbox.
+     * @param {Ext.form.field.Checkbox} cb
+     * @param {boolean} checked
+     */
+    onCheckChange: function(cb, checked) {
+        var slider = cb.up().down('multislider');
+        if(slider) {
+            slider.setDisabled(!checked);
+        }
+        if(checked) {
+            this.applySliderEffects();
+        } else {
+            this.removeSliderEffects();
+        }
+    },
+
+    /**
+     * Applies the effect of the slider, e.g. filters by the numeric values.
+     */
+    applySliderEffects: function() {
+        var me = this;
+        var view = me.getView();
+        var slider = view.down('multislider');
+        var orderedVals = me.orderValues(slider.getValues());
+        var layers = me.getAffectedLayers();
+        var fieldName = view.getNumericField();
+        // between-filter is also inclusive
+        var gteFilter = GeoExt.util.OGCFilter.getOgcFilter(
+            fieldName, 'gte', orderedVals[0], '2.0.0'
+        );
+        var lteFilter = GeoExt.util.OGCFilter.getOgcFilter(
+            fieldName, 'lte', orderedVals[1], '2.0.0'
+        );
+        var combined = GeoExt.util.OGCFilter.combineFilters(
+            [gteFilter, lteFilter], 'And', false
+        );
+
+        Ext.each(layers, function(layer) {
+            var source = layer.getSource();
+            if (source instanceof ol.source.Cluster) {
+                source = source.getSource();
+            }
+            if (layer.get('isWfs')) {
+                source.set('numericFilters', combined);
+                source.clear();
+                source.refresh();
+            } else if (layer.get('isWms')) {
+                // What about existing FILTER in params?
+                // same goes for the case when we remove the filter, see below.
+                // …Replace any existing filters for now
+                source.updateParams({
+                    'FILTER': combined
+                });
+            }
+        });
+    },
+
+    /**
+     * Removes the effects of the slider, e.g. unfilters by the numeric values.
+     */
+    removeSliderEffects: function () {
+        this.getAffectedLayers();
+        var layers = this.getAffectedLayers();
+        Ext.each(layers, function(layer) {
+            var source = layer.getSource();
+            if (source instanceof ol.source.Cluster) {
+                source = source.getSource();
+            }
+            if (layer.get('isWfs')) {
+                source.set('numericFilters', null);
+                source.clear();
+                source.refresh();
+            } else if (layer.get('isWms')) {
+                // …unset any existing filters for now
+                source.updateParams({
+                    'FILTER': undefined
+                });
+            }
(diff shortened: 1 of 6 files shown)


marcjansen:
Aybody up for a review?

app/controller/panel/NumericAttributeSlider.js
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * Returns those layers that the slider will affect.
+     */
+    getAffectedLayers: function() {
+        return BasiGX.util.Layer.getLayersBy('isNumericDependent', true);
+    },
+
+    /**
+     * Orders the values of the slider so that the first index always is the
+     * lower value, and the secind is at least equally big.
```
chrismayer: Typo => `second`

app/factory/Layer.js
```diff
@@ -240,10 +242,15 @@ Ext.define('CpsiMapview.factory.Layer', {
             // this within the function is bound to the vector source it's
             // called from.
             var timeFilters = this.get('timeFilters');
-
             if (!Ext.isEmpty(timeFilters)) {
                 allFilters = Ext.Array.merge(allFilters, timeFilters);
             }
+            var numericFilters = this.get('numericFilters'); // TODO docs
```
chrismayer: Anything up to come since there are several `TODO` tags in the following lines? Or just a reminder for future tasks?

app/factory/Layer.js
```diff
@@ -325,7 +332,11 @@ Ext.define('CpsiMapview.factory.Layer', {
             isTimeDependent: !!layerConf.timeitem,
             dateFormat: layerConf.dateFormat,
             timeProperty: layerConf.timeitem,
-            isWfs: true
+            isWfs: true,
+            // TODO docs
+            // TODO wouldn't it make sense to have the actual field here
+            //      instead of at the slider and globally for all layers?
```
chrismayer: Totally agreed. It would be better to have the field in here and detect `isNumericDependent` based on the existence of this field.

app/view/panel/NumericAttributeSlider.js
```diff
@@ -0,0 +1,132 @@
...
+     * @param {boolean} newDisabled
+     */
+    setDisabled: function(newDisabled) {
+        // override the original method and
+        // fast forward to checkbox check / uncheck
+        var newEnabled = !newDisabled;
+        this.setEnabled(newEnabled);
+        var cb = this.down('checkbox[name="active"]');
+        cb.setValue(newEnabled);
+        return this; // the original methoid is chainable
```
chrismayer: Typo `method`

app/view/toolbar/MapTools.js
```diff
@@ -84,6 +86,14 @@ Ext.define('CpsiMapview.view.toolbar.MapTools', {
             xtype: 'cmv_timeslider',
             startDate: new Date(2014, 0, 1),
             endDate: new Date(2020, 11, 30)
+        }, {
+            xtype: 'cmv_numericattributeslider',
+            numericField: 'Speed_Limit', // TODO might be better suited at thelayer level
```
chrismayer: Typo missing space `the layer`

app/view/panel/NumericAttributeSlider.js
```diff
@@ -0,0 +1,132 @@
...
+    initComponent: function () {
+        var me = this;
+        var values = [me.getCurrLowerValue(), me.getCurrUpperValue()];
+        me.items = [{
+            xtype: 'multislider',
+            name: 'slider',
+            labelAlign: 'right',
+            listeners: {
+                afterrender: 'initSlider',
+                changecomplete: 'applySliderEffects'
```
chrismayer: Do I get it right that the `applySliderEffects` is always bound to `changecomplete` but the slider gets disabled in case the checkbox is unchecked?

app/factory/Layer.js
marcjansen: I am not sure where I would document this. Do we have one place where the layer properties we use internally are documented? Likewise, for the initial JSON where is the one true documentation for the structure and allowed values?

app/factory/Layer.js
marcjansen: This something that @geographika has to say something to, I think. This can be changed afterwards in any way we want it.

app/view/panel/NumericAttributeSlider.js
marcjansen: Yes. the slider gets disabled according to the checkbox state and depending on the state we also apply / revert the effects of the thumbs. On might argue that disabling this parent component would completely disable the checkbox as well. This can be changed if we want. Does that answer your question?

app/factory/Layer.js
geographika: I'm have a drafy JSONSChema doc for the JSON config at http://w08-mapserver.compass.ie/mapview-config-docs/. I'll move it into this project in the future, At the moment the config is very flexible and driven more by the feature requirements than a config spec, so any changes to the config are fine.

app/factory/Layer.js
geographika: Agree far more flexible to have the field set at the layer level rather than the component.

app/view/panel/NumericAttributeSlider.js
chrismayer: Yes, this answers my question. All fine from my side :+1:

app/factory/Layer.js
marcjansen: I'd like to do this in a follow-up PR if that is fine for you.

chrismayer: Fine for me. Maybe we open an issue for that, so we keep it in mind?

marcjansen: Please see here for my thoughts on the topic: https://github.com/meggsimum/cpsi-mapview/issues/99

marcjansen:
How do we proceed from here on? I suggest merging and then changing implementation details in another PR in a future sprint.

geographika:
> How do we proceed from here on? I suggest merging and then changing implementation details in another PR in a future sprint.

Sounds good to me.

marcjansen:
More discussion is in #99.

PR #96, opened 2019-05-28
Render SV panorama to body of floating window

chrismayer:
This PR ensures that the SV panorama is rendered to the body of the floating window. So the layout issue described in #71 gets fixed.


chrismayer:
Thanks for the review @annarieger!

PR #113, opened 2019-08-19
Correct URL for ESRI REST World_Light_Gray_Base

chrismayer:
This corrects the URL of the ESRI REST World_Light_Gray_Base layer since the x-y-z declaration for the tiles has to be given in the config URL (introduced with #107).


chrismayer:
Thx @ahennr

PR #114, opened 2019-08-21
Tree plugin for radio buttons to switch layer styles

### LINKED ISSUE #111: Create radio buttons for layer themes
The old version of mapview had a class to generate radio buttons for an array of styles associated with a WFS layer. 

radio_sld

This allowed switching between different SLD files for the same vector features - [code here](https://github.com/compassinformatics/mapview/blob/master/src/gis/tree/CustomWfsLoader.js)

SLD file names were stored in the layer config file as a comma delimtied list e.g. `"slds": "Collision Type,Accident Type"` in the JSON config file. 
			
			
It would be good to apply similar functionality in the new version of mapview. The feature would also benefit from being compatible with both WFS and WMS. 
For WMS layers the selected STYLE would be sent to the WMS server. 
For WFS the selected SLD file would be applied. Code to apply SLDs to layers is currently [in the LayerFactory](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/factory/Layer.js#L354)

**Note** - if vector tiles work well (see #102) then there is no immediate requirement to support WFS layers. If vector tiles can support 1,000s of features then it may make more sense to allow users to switch styles on vector layers directly?

The config could be a proper list rather than a string e.g. `"styles": ["style1","style2"]`

The same config could be used for WFS referencing SLD files: `"styles": ["style1.xml","style2.xml"]`

The method used to apply the styles must be compatible with `CpsiMapview.view.menuitem.LayerLabels` which adds a label style on top of any existing style.

chrismayer:
This PR introduces a plugin (`cmv_tree_column_style_switcher`) for the `TreeColumn` of the `LayerTree` in order to render radio buttons to switch the layer style.

By setting a property `styles` in a WMS or WFS configuration the feature can be activated:

```json
{
    "layerType": "wms",
    ...
    "styles": ["My Style 1", "My Style 2"],
    ...
}
```
```json
{
    "layerType": "wfs",
    ...
      "stylesBaseUrl": "https://my-super-url/to/sld-files/",
      "styles": ["My_Style_1.xml", "My_Style_2.xml"],
    ...
}
```
 The styles will be shown as radio buttons at the corresponding `LayerTree` node.

grafik

**Side note:** In case you want to ensure that the filter values of the loaded SLD (for WFS rendering) are interpreted as number instead of strings (where possible) you can set the config property `stylesForceNumericFilterVals` on the layer:

```json
{
    "layerType": "wfs",
    ...
      "stylesBaseUrl": "https://my-super-url/to/sld-files/",
      "styles": ["My_Style_1.xml", "My_Style_2.xml"],
      "stylesForceNumericFilterVals": true
    ...
}
```
This ensures that a filter `["==", "myProp", '1']` will be transformed to `["==", "myProp", 1]`. Necessary since SLD / FE is not aware of data types for `PropertyIsEqual` filter.

Closes #111


app/plugin/TreeColumnStyleSwitcher.js
```diff
@@ -0,0 +1,162 @@
+/**
+ * A plugin for Ext.grid.column.Column that extends the internal cellTpl to
+ * add a radio group to switch the style for a layer
+ * (CpsiMapview.view.layer.StyleSwitcherRadioGroup).
```
ahennr:
```suggestion
 * (CpsiMapview.plugin.TreeColumnStyleSwitcher).
```

chrismayer: This is meant as reference for class offering the RadioGroup mentioned in the text.

chrismayer:
Thanks for the review @ahennr. I'll merge now.

Issue #105, opened 2019-07-25
Add Karma testing to the project


geographika:
Linting is already run on the project as part of CI with Travis and locally with NodeJS. 

Ideally some sample tests will be added using Karma (the suite used in GeoExt), along with all necessary changes to package.json to run these from the command line with Node. 
Some "best practices" on how/what to test are welcomed. 

GeoExt can be used as a template for the testing/linting approach, although only tests and linting are required currently (no need for docs building etc. yet). 

I've gone through the docs at https://github.com/geoext/geoext3/blob/master/development.md which are very useful in going throught the steps on testing locally. 

In addition it would be useful to setup testing for https://github.com/geographika/geoext-demo-app - this is more of an application rather than a library of components, so would be very useful how a GeoExt-based application should be tested.

verendi:
Priority is on Cpsi-mapview

ahennr:
https://github.com/geographika/geoext-demo-app/pull/1

ahennr:
https://github.com/compassinformatics/cpsi-mapview/pull/121

https://github.com/compassinformatics/cpsi-mapview/pull/122

geographika:
Thanks for these. I'll be using them as a template for other projects.

PR #115, opened 2019-08-21
Introduce separate lint script and upgrade eslint 

chrismayer:
This PR does 2 things:

  - Introduce separate `lint` script (besides calling it in `test` script)
  - Upgrade eslint to v6.2.1

First small preparation steps for #105


chrismayer:
Thanks @ahennr!

PR #116, opened 2019-08-22
Spatial query tool

ahennr:
Introduces a new component `SpatialQueryButton` that uses `GeoExt.util.OGCFilter` to query a configured/associated layer with a spatial filter. A polygon, a linestring or a point can be drawn to define the geometry used in the filter.

Depends on https://github.com/geoext/geoext3/pull/566/ and https://github.com/terrestris/BasiGX/pull/497

Please note, that that queried WFS has CORS enabled or use NGinx proxy. 
In addition, the layer factory was enhanced to support  required fields for the name of geoemtry field and namespace definitions needed for WFS `GetFeature`.

```
"geomFieldName": "geometry",
"namespaceDefinitions": {
    "osm": "http://terrestris"
}
```

Resolves https://github.com/compassinformatics/cpsi-mapview/issues/101

PR #119, opened 2019-09-02
Switch to getting mapComponent

geographika:
When testing the new SpatialQueryButtonController in #116 I was getting errors that `view.map is null` on the `view.map.addInteraction(me.drawQueryInteraction);` line. 

Also the code was getting a map rather than a mapComponent with:

`var mapComp = CpsiMapview.view.main.Map.guess();`

This switches it to:

`var mapComp = me.mapComponent || BasiGX.util.Map.getMapComponent();`

If not `mapComp.getMap()` is `undefined. 

Several other tools use the same approach e.g. the [DigitizeButtonController](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/button/DigitizeButtonController.js)

@ahennr - could you check the changes to see that they make sense?
Thanks for all the work on the tool!

### DIFF
--- a/app/controller/button/SpatialQueryButton.js
+++ b/app/controller/button/SpatialQueryButton.js
@@ -102,7 +102,9 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
             return;
         }
 
-        var mapComp = me.mapComponent || BasiGX.util.Map.getMapComponent();
+        if (!me.mapComponent) {
+            me.mapComponent = BasiGX.util.Map.getMapComponent();
+        }
 
         var projString = mapComp.getMap().getView().getProjection().getCode();
         var geomFieldName = view.queryLayer.get('geomFieldName') ||
--- a/app/controller/button/SpatialQueryButton.js
+++ b/app/controller/button/SpatialQueryButton.js
@@ -13,6 +13,16 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
     drawQueryInteraction: null,
 
     /**
+    * The OpenLayers map. If not given, will be auto-detected
+    */
+    map: null,
+
+    /**
+     * The BasiGX mapComponent. If not given, will be auto-detected
+     */
+    mapComponent: null,
+
+    /**
      * Function to determine the query layer if not yet defined in class
      */
     findQueryLayer: function () {
@@ -35,6 +45,13 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
         var me = this;
         var view = me.getView();
 
+        if (view.map && view.map instanceof ol.Map) {
+            me.map = view.map;
+        } else {
+            // guess map as fallback
+            me.map = BasiGX.util.Map.getMapComponent().map;
+        }
+
         if (!view.queryLayer) {
             me.findQueryLayer();
         }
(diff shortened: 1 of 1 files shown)


app/controller/button/SpatialQueryButton.js
```diff
@@ -84,7 +101,9 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
         if (!view.queryLayer) {
             return;
         }
-        var mapComp = CpsiMapview.view.main.Map.guess();
+
+        var mapComp = me.mapComponent || BasiGX.util.Map.getMapComponent();
```
ahennr:
```suggestion
        if (!me.mapComponent) {
            me.mapComponent = BasiGX.util.Map.getMapComponent();
        }
```

ahennr: As `mapComponent` is not set currently, I would prefer to to set `me.mapComponent` where and use `me.mapComponent` afterwards, what to you think?

app/controller/button/SpatialQueryButton.js
```diff
@@ -128,7 +147,7 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
     onWfsExecuteSuccess: function (response) {
         var me = this;
         var view = me.getView();
-        var mapComp = CpsiMapview.view.main.Map.guess();
+        var mapComp = me.mapComponent || BasiGX.util.Map.getMapComponent();
```
ahennr: In #116, @chrismayer asked me to use `CpsiMapview.view.main.Map.guess();` in favor of `BasiGX.util.Map.getMapComponent();` since this detects the cpsi map-panel directly by it's `xtype`. Maybe the guess function should be enhanced to allow a "fuzzy" search if no `cmv_map` was found. What do you think @geographika @chrismayer  ?

geographika: Ah ok I thought that function was searching for a map rather than mapComponent. I was running with the latest cpsi-mapview project and getting an error when loading the application - you didn't encounter this?

chrismayer: I think the clear way is to inject the map component when creating SpatialQueryButton component. The guess methods should only be used as fallback.
Nevertheless the guess function of BasiGX seems more flexible and thus it is maybe better to use it here.

app/controller/button/SpatialQueryButton.js
geographika: Yes that makes sense.
The [DigitizeButton](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/button/DigitizeButtonController.js#L289) currently uses:

`var mapComponent = me.mapComponent || BasiGX.util.Map.getMapComponent();`

It would make sense to harmonize the approach for all tools (also with regards to `CpsiMapview.view.main.Map.guess()`)

ahennr: :+1:

app/controller/button/SpatialQueryButton.js
ahennr: > Ah ok I thought that function was searching for a map rather than mapComponent. I was running with the latest cpsi-mapview project and getting an error when loading the application - you didn't encounter this?

@geographika : I successfully checked the tool in cpsi-mapview as well, but maybe in an older version than you. However, could you please so kind to include the suggestion of @chrismayer of passing `map` / `mapComponent` to `SpatialQueryButton`?

geographika:
@ahennr - is this ok to merge now? I have a follow up pull request which adds a new event to this class to help integrate with a full application (which will set the map directly with a controller).

ahennr:
@geographika feel free to merge :+1: :smiley: 

Regarding CORS: AFAIK CORS is already enabled for https://ows.terrestris.de/geoserver/osm/wfs, see screenshot below…

image

Issue #112, opened 2019-07-31
Update layer record UI in tree when outside visible scale


geographika:
The old mapview appended scale text to tooltips to show the user when a map layer was out of range [implemented here](https://github.com/compassinformatics/mapview/blob/39616abc8b5e2f5a73e36ee4048665ffc835d09e/src/gis/tree/LayerNode.js#L95), and also changed the layer text to italics:

![layer_scale](https://user-images.githubusercontent.com/490840/62229112-48062b00-b3bf-11e9-9b48-1efbdffff2c5.png)

This used the [LayerNode](https://geoext.github.io/geoext2/docs/#!/api/GeoExt.tree.LayerNode) functionlity in GeoExt2. There is currently no equivalent in GeoExt3. 

Layer scales can be set in the JSON config file (and are already handled in the [LayerFactory](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/factory/Layer.js#L562).

geographika:
@ahennr - thanks for this - looks great. 
Note to myself -add the following to app.json to include cpsi-mapview CSS in a parent project:

PR #123, opened 2019-09-04
Layer tree - Update tree nodes depending on layers resolution restrictions

ahennr:
This PR adds a new plugin `TreeColumnInResolutionRange` that changes opacity and font style of a layer tree node if the assigned layer is not in range. In addition, the visible scales are shown in a qtip if tree node is hovered.  

scale_dependent_nodes

Depends on https://github.com/geoext/geoext3/pull/580, as soon as this is merged, I'll remove WIP state

### DIFF
--- a/app/plugin/TreeColumnInResolutionRange.js
+++ b/app/plugin/TreeColumnInResolutionRange.js
@@ -137,7 +137,7 @@ Ext.define('CpsiMapview.plugin.TreeColumnInResolutionRange', {
         }
 
         if (maxScale && minScale) {
-            scale = Ext.String.format('Visible between <b>1:{0}</b> and <b>1:{1}</b>', maxScale, minScale);
+            scale = Ext.String.format('Visible between <b>1:{0}</b> and <b>1:{1}</b>', minScale, maxScale);
         } else {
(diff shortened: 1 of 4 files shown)


ahennr:
Closes https://github.com/compassinformatics/cpsi-mapview/issues/112

app/plugin/TreeColumnInResolutionRange.js
```diff
@@ -0,0 +1,157 @@
...
+        if (!Ext.isDefined(node.originalQtip)) {
+            node.originalQtip = node.getOlLayer().get('description') || '';
+        }
+
+        if (inRange) {
+            return node.originalQtip.length > 0 ? node.originalQtip : undefined;
+        }
+
+        var currentTip = node.getOlLayer().get('description');
+        if (currentTip && (currentTip.indexOf('<br />') > -1 ||
```
chrismayer: Maybe a more robust selector than `'<br />'` would be good here.

ahennr: See new issue https://github.com/compassinformatics/cpsi-mapview/issues/124

PR #126, opened 2019-09-05
Fire an event with the spatial Ext.util.Filter

geographika:
Trigger a new custom event containing the spatial filter created by the tool. This can then be combined with other filters external to the tool to make complex filters. 

The current functionality is preserved by default, but can be deactivated by setting the triggerWfsRequest flag to false in cases where the button should simply create a filter and fire the event.

### DIFF
--- a/app/controller/button/SpatialQueryButton.js
+++ b/app/controller/button/SpatialQueryButton.js
@@ -23,6 +23,13 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
     mapComponent: null,
 
     /**
+    * If set to true a Wfs GetFeatures request will be automatically triggered
+    *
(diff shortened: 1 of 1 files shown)


app/controller/button/SpatialQueryButton.js
```diff
@@ -80,14 +87,52 @@ Ext.define('CpsiMapview.controller.button.SpatialQueryButtonController', {
         }
     },
 
+    /**
+    * Normalizes the given degree value.
```
ahennr: IMHO the method description is misleading. Could you please provide a more precise one?

geographika: @ahennr - thanks for spotting this! It was a copy/paste from a different function..

PR #130, opened 2019-09-06
Add jsonix to js dependencies

ahennr:
Title says it all.


chrismayer:
Thanks @ahennr. LGTM, please merge.

Do you know if it possible to add JSONIX via npm? Maybe this would be another option to embed it (Just an option - nothing to do here).

Issue #103, opened 2019-07-25
Add Switch Layer type to the LayerFactory


geographika:
The original mapview project has a [Switch Layer type](https://github.com/compassinformatics/mapview/blob/master/src/gis/tree/SwitchLayerNode.js). This uses a single tree node in the layer tree, but switches from WMS to WFS when zoomed in on the map.
This allows a seamless transition between the two types, allowing tooltips at the lower level, but for data to be displayed a large extents. 

TODISCUSS Maybe vector tiles may make this approach redundant.

PR #131, opened 2019-09-20
Add switch layer functionality

JakobMiksch:
Hey there!
I created a [switch layer](https://github.com/compassinformatics/cpsi-mapview/issues/103) type in the layer factory. 

It adds a listener on the resolution: when the resolution changes all layers are checked if they are switch layers and if they need to be switched. After a switch the layertree gets updated.

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -108,87 +108,6 @@ Ext.define('CpsiMapview.factory.Layer', {
     },
 
     /**
-     * Loops through all layers, identifies switch layers
-     * and replaces them if required
-     *
-     * @param {ol.Object.Event} evt The event which contains the view.
-     */
-    handleSwitchLayerOnResolutionChange: function (evt) {
-        var resolution = evt.target.getResolution();
-        var allLayers = BasiGX.util.Map.getMapComponent().getMap().getLayers();
-        var overlayGroup = BasiGX.util.Layer.getLayerByName('Layers', allLayers);
-
-        if (!Ext.isDefined(overlayGroup)) {
-            return;
-        }
-        var overlayCollection = overlayGroup.getLayers();
-
-        overlayCollection.forEach(function (layer, index) {
-
-            if(layer.get('isSwitchLayer') && LayerFactory.isLayerSwitchNecessary(layer, resolution)){
-
-                var switchConfiguration = layer.get('switchConfiguration');
-                var newLayer = LayerFactory.createSwitchLayer(switchConfiguration);
-
-                overlayCollection.setAt(index, newLayer);
-
-                LayerFactory.updateLayerTreeForSwitchLayers();
-            }
-        });
-    },
-
-    /**
-     * Checks if the switch layer has to be replaced
-     *
-     * @param {*} layer       the layer to check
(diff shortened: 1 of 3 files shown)


app/factory/Layer.js
```diff
@@ -78,6 +78,13 @@ Ext.define('CpsiMapview.factory.Layer', {
             //do nothing, and return empty layer
         }
 
+        // TODO: make sure this listener is called only once
+
+        // listener that checks the resolution and
+        // changes the switch layer if required
+        var olMap = BasiGX.util.Map.getMapComponent().getMap();
+        olMap.getView().on('change:resolution', LayerFactory.handleSwitchLayerOnResolutionChange);
```
JakobMiksch: I am not sure if this listener gets created once or multiple times. The performance seems to fine. But it would be cleaner to make sure that listener is created only once.

app/factory/Layer.js
```diff
@@ -107,6 +114,87 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+     *
+     * @param {ol.Object.Event} evt The event which contains the view.
+     */
+    handleSwitchLayerOnResolutionChange: function (evt) {
+
+        var resolution = evt.target.getResolution();
+
+        // TODO: this needs to be called more elegantly
+        var allLayers = BasiGX.util.Map.getMapComponent().getMap().getLayers();
+        var overlayGroup =  BasiGX.util.Layer.getLayerByName('Layers', allLayers).getLayers();
```
JakobMiksch: Is there a better to way to call the overlay group?

app/factory/Layer.js
```diff
@@ -107,6 +114,87 @@ Ext.define('CpsiMapview.factory.Layer', {
...
+
+            // only change for switch layers
+            if(switchConf){
+                node.triggerUIUpdate();
+            }
+        });
+
+        // TODO: not sure if this is required (?)
+        if(treePanel) {
+            treePanel.fireEvent('itemupdate');
```
JakobMiksch: I basically adapted the code for the layer tree update from the "LayerTreePlugin". The update works without firing the `itemupdate` event. Not sure if this is really required.

chrismayer:
Thanks @JakobMiksch for this. Looks good so far to me. I'll have a look at the issues / remarks you added and will come up with some adaptions (if needed).

app/factory/Layer.js
chrismayer: Yes it is bound multiple times, I'll come up with a change for that.

app/factory/Layer.js
chrismayer: At the moment there is no better way. Maybe we should introduce a util class to query the layer groups in the future.

app/factory/Layer.js
chrismayer: Firing 'itemupdate' is not necessary. Removed in f880373.

app/factory/Layer.js
chrismayer: See cf91b4c

chrismayer:
Thanks for the review @ahennr and thanks again for your work @JakobMiksch! I'll merge now.

Issue #117, opened 2019-08-23
Switch Layer Legend with Style


geographika:
As referred to at #111 

Is it possible to update the associated layer legend with the selected style? The legend would update each time a radio style button is selected.  

Currently the layer factory uses a simplistic legendUrl with a hard-coded style. 
This is then used by `CpsiMapview.plugin.TreeColumnStyleSwitcher` to set the legend in the tree. 

For WMS layers the GetLegendGraphic URL could be dynamically generated based on the WMS URL, parameters, and selected style. For MapServer at least the same WMS GetLegendGraphic URL would also work for WFS legends.

![layertree](https://user-images.githubusercontent.com/490840/63576921-87263700-c58d-11e9-9993-06f2e240061f.png)

chrismayer:
> For MapServer at least the same WMS GetLegendGraphic URL would also work for WFS legends.

Could you provide an example for a WFS GetLegendGraphic request in MapServer?

https://plmonaghandev.compass.ie/mapserver/?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetLegendGraphic&FORMAT=image%2Fpng&TRANSPARENT=TRUE&SLD_VERSION=1.1.0&LAYER=LightUnit&STYLE=LightUnit_Unit_Type.xml

brings `msWFSDispatch(): WFS server error. Invalid WFS request: GetLegendGraphic`

geographika:
I can use a naming convention for the XML style exports and the style name in this case LAYERNAME_STYLENAME 
I previously replaced any spaces in the stylename with underscores - but I can work around this.

chrismayer:
> I previously replaced any spaces in the stylename with underscores - but I can work around this.

Oh if that is the convention I can work with that. So for WMS

LAYERS: `LightUnit` 

and

STYLE:  `Unit Types` 

become 

`LightUnit_Unit_Type.xml` (`LAYERS_STYLE<with underscores>.xml`) for WFS, correct? If yes I can implement this already.

geographika:
@chrismayer - yes that is the convention I was using. Good news!

geographika:
@chrismayer - very nice!
I have one question -  why do the layers require a legendWidth or legendHeight defined for the layer?
I tried without and there are some very cryptic ExtJS errors and requests to undefined URLs, so I know it must be fixing this, but not sure why.

chrismayer:
[...]

PR #132, opened 2019-09-25
Switch layer legend with style

chrismayer:
This ensures that the legend for WMS / WFS is shown by `getLegendGraphic` request (if not defined as static image) and that the legend will be updated in case the layer style is changed by the user.

Convention to activate the automatic `getLegendGraphic` is that either `legendWidth` or `legendHeight` has to be defined in the WMS / WFS layer confinguration and no property `legendUrl` is set.

For WFS the WMS `getLegendGraphic` request is used by convetion, as specified in #117.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,7 @@ module.exports = {
         "GeoExt": false,
         "CpsiMapview": false,
         "LayerFactory": false,
+        "LegendUtil": false,
         "GeoStylerSLDParser": false,
         "GeoStylerOpenlayersParser": false,
         "google": false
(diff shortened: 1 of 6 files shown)


chrismayer:
Thanks for the review @ahennr! Will merge now.

geographika:
@chrismayer - is the pull request missing a file? It requires `CpsiMapview.util.Legend` but this file/namespace doesn't exist.

chrismayer:
Yes, it misses a file. Sorry, stupid me! #133 should fix it. Are you available for a short review @geographika? TIA!

PR #137, opened 2019-09-30
Reuse Style Name function

geographika:
Reuse LegendUtil.getWmsStyleFromSldFile
Fix a comment typo

### DIFF
--- a/app/view/layer/StyleSwitcherRadioGroup.js
+++ b/app/view/layer/StyleSwitcherRadioGroup.js
@@ -68,7 +68,8 @@ Ext.define('CpsiMapview.view.layer.StyleSwitcherRadioGroup', {
     getLayerStyleLabel: function (layerStyle) {
         if (this.layer.get('isWfs') || this.layer.get('isVt')) {
             // remove _ and the .xml file ending
-            return LegendUtil.getWmsStyleFromSldFile(layerStyle);
+            var legendUtil = CpsiMapview.util.Legend;
+            return legendUtil.getWmsStyleFromSldFile(layerStyle);
         } else {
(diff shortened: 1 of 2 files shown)


app/view/layer/StyleSwitcherRadioGroup.js
```diff
@@ -66,7 +68,7 @@ Ext.define('CpsiMapview.view.layer.StyleSwitcherRadioGroup', {
     getLayerStyleLabel: function (layerStyle) {
         if (this.layer.get('isWfs') || this.layer.get('isVt')) {
             // remove _ and the .xml file ending
-            return layerStyle.replace(/_/g, ' ').replace('.xml', '');
+            return LegendUtil.getWmsStyleFromSldFile(layerStyle);
```
marcjansen: Where is the definition for `LegendUtil`? I think sth. like

```js
var LegendUtil =  CpsiMapview.util.Legend;
```

is missing

marcjansen: Yeah, this will then be available directly. Unsure whether this is something that I **like**, but it should  at least work :smiley:

geographika: Agree - will keep it explicit!

chrismayer: Just to clarify: Yes `LegendUtil`, is defined as `alternateClassName` in `CpsiMapview.util.Legend`. By adding `'CpsiMapview.util.Legend'` to the `requires` option (what you have to do in any case) it will be resolved. This is a quite common approach in ExtJS to have short aliases for singletons.

PR #141, opened 2019-10-09
Add generic base classes for WFS grids

geographika:
This pull request contains the following:

A view, view model, and controller for a generic WFS grid that can be reused in several applications. It contains attribute and spatial filtering, exports, paging and sorting. 

A new layer plug-in is also included to open a WFS grid associated with the layer.

Some minor edits have been made to the SpatialQueryButton to allow the layer to be set from the grid configuration. 

An example grid configuration with the model, store, viewmode, and grid view all in one file has been added `CpsiMapview.view.grid.ExampleGrid` to allow a demo in the cpsi-mapview project. 

A screencase showing the grid layer-plugin and the grid view with a linked spatial query tool is [available here](https://drive.google.com/file/d/1_euqALWDjRpLzv9Bz-kLrcJl5ojC-XH6/view).


app/view/grid/ExampleGrid.js
```diff
@@ -0,0 +1,76 @@
...
+        vectorLayerKey: 'LIGHT_UNIT_MVT',
+        gridStoreType: 'GridExample',
+        gridLayerName: 'GridExampleLayer' // TODO this is duplicated in layerOptions above
+    }
+});
+
+// create the grid view
+Ext.define('CpsiMapview.view.grid.ExampleGrid', {
+    extend: 'CpsiMapview.view.grid.Grid',
+    xtype: 'examplegrid',
```
chrismayer: A `cmv_` would be good here to keep the xtypes consistent

app/view/grid/Grid.js
```diff
@@ -0,0 +1,105 @@
...
+    width: 1050,
+
+    bind: {
+        store: '{gridstore}'
+    },
+
+    viewConfig: {  //this config is passed to the view
+        loadMask: {
+            msg: 'Loading records' // TODO not sure why this isn't applied to the loadMask
+        }
```
chrismayer: Could be achieved with `loadingText: 'Loading records'`, see https://docs.sencha.com/extjs/6.2.0/classic/Ext.view.Table.html#cfg-loadingText

app/view/menuitem/LayerGrid.js
```diff
@@ -0,0 +1,80 @@
...
+     * create a new grid and associated window.
+     */
+    handlerFunc: function () {
+        var me = this;
+
+        var gridXType = me.layer.get('gridXType');
+        var title = me.layer.get('name');
+
+        // we can't keep a reference to the window in this class
+        // as a new Ext.menu.Item i created each time the menu is
```
chrismayer: typo ` ... is created ...`

app/controller/grid/Grid.js
```diff
@@ -0,0 +1,328 @@
...
+     * @param {ol.Feature} feature
+     * @param {ol.Map} map
+     * @private
+     */
+    zoomToFeature: function (feature, map) {
+
+        // TODO check for feature type when zooming
+        var extent = feature.getGeometry().getExtent();
+        // as this is a point then buffer it by 100m
+        extent = ol.extent.buffer(extent, 100);
```
chrismayer: Making the buffer width configurable in the grid would be a good idea, IMHO.

app/controller/grid/Grid.js
```diff
@@ -0,0 +1,328 @@
...
+        var me = this;
+        var map = me.getViewModel().get('map');
+
+        if (!me.contextMenu) {
+            me.contextMenu = Ext.create('Ext.menu.Menu', {
+                items: [{
+                    text: 'Zoom to Feature',
+                    scope: me,
+                    handler: function () {
+                        var record = grid.getSelection()[0];
```
chrismayer: This should be the `record` variable provided in the `onItemContextMenu` listener function.

app/controller/grid/Grid.js
```diff
@@ -0,0 +1,328 @@
...
+            return;
+        }
+
+        grid.setLoading('Exporting to Excel...');
+
+        // later in an event listeners
+        grid.saveDocumentAs({
+            type: 'xlsx',
+            title: 'Map Export',
+            fileName: 'myExport.xlsx'
```
chrismayer: Would be good to have these properties as config in the grid view class

app/controller/grid/Grid.js
```diff
@@ -0,0 +1,328 @@
...
+    *
+    * @cfg {Ext.util.Filter} spatialFilter
+    */
+    spatialFilter: null,
+
+    /**
+     * Fire a onRowDblClick to be handled by the parent application
+     */
+    onRowDblClick: function (sender, record) {
+        this.fireEvent('onRowDblClick', sender, record);
```
chrismayer: Is it really necessary to forward the `rowdblclick` event of the grid?

app/view/grid/Grid.js
geographika: I've tried various properties such as `loadingText `, `loadMessage` in and out of the `viewConfig` with no luck. 
I can set it with `view.setLoading('Loading Records...');` in `onWfsStoreBeforeLoad` with no problems though.

app/controller/grid/Grid.js
geographika: In a child application I used this approach, but I guess I can listen to the event directly from a view without needing to forward anything. Will remove from here.

app/controller/grid/Grid.js
geographika: The record always seems to be available when `onItemContextMenu` is called but is then undefined in the handler. I can set me.record in the `onItemContextMenu` function and access this in the handler, but this doesn't seem the cleanest approach. I'm unsure why the variable isn't available in the anonymous function.

geographika: Updated

app/controller/grid/Grid.js
geographika: Added as config option

app/view/menuitem/LayerGrid.js
geographika: Fixed

app/view/grid/ExampleGrid.js
geographika: Updated

app/controller/grid/Grid.js
geographika: Added

geographika:
@chrismayer - many thanks for the review! I've addressed the issues. There are likely a few more options that will become configurable over time as this base grid is used in more applications. 
The `loadingText` remains a mystery to me, but currently a minor display issue. 
I've fixed the spatial filter paging bug - thanks for catching that. 
Will merge tomorrow.

chrismayer:
Thanks for your ongoing effort @geographika, very nice.

Regarding to the `loadingText`: Agree, if it does not work as described in the API-Docs it is maybe a bug in ExtJS. So if `view.setLoading('Loading Records...')` works then it is fine of course.

Issue #146, opened 2019-10-14
Add Layer Transparency Menu Item


geographika:
Add a new layer menu item, similar to `CpsiMapview.view.menuitem.LayerRefresh` that has a transparency slider that can be used to apply transparency to the layer. 

![image](https://user-images.githubusercontent.com/490840/66757420-71056a80-ee9c-11e9-9698-746261bcd8d4.png)

![image](https://user-images.githubusercontent.com/490840/66757491-998d6480-ee9c-11e9-937f-ee270f9d4330.png)

geographika:
@chrismayer - thanks - works very well.

PR #148, opened 2019-10-21
Add layer opacity slider context menu 

chrismayer:
This adds a layer opacity slider context menu for all layers in the tree (see #146). 

grafik

Add `opacitySlider: false` in the layer configuration to avoid the layer opacity slider.

NOTE: Since using `setOpacity` for `ol.layer.VectorTile` instances has no effect, this introduces a workaround by changing `globalAlpha` in `precompose` of the layer.
See openlayers/openlayers#4758


geographika:
@chrismayer - thanks for this, looks great.

PR #164, opened 2020-01-27
Fix broken layer configuration

chrismayer:
This PR fixes the broken syntax of the layers configuration and corrects some indentation.


marcjansen:
Looks good to me. We should enforce validity with a linter. A PR adding this would be nice.

chrismayer:
Thanks for teh review @marcjansen! You are right we should add a linter. I added an issue as reminder for this.

geographika:
Thanks @chrismayer - changes look good.

PR #174, opened 2020-02-12
bug in error message display fixed

leonardospina:
the error message wasn't actually displayed because of the missing brackets


geographika:
Thanks @leonardospina !

PR #175, opened 2020-02-17
Fix disappearing style radios when moving layer

### LINKED ISSUE #169: Radio buttons disappear when moving a layer in the layer list
When dragging a layer with multiple styles and a radio button UI to a different location in the layer list, the radio buttons disappear:

Before move:

image

After move:

image

chrismayer:
This ensures that the style radio groups in the LayerTree are rendered after a node has been dragged and dropped. Otherwise they disappear.

Fixes #169


chrismayer:
Thanks for the review @jansule! I'll merge now...

Issue #171, opened 2020-02-06
Circle Selection Tool


geographika:
Create a new circle drawing tool, that allows a user to click a point and drag outwards to make the circle larger. The same functionality as demoed at https://openlayers.org/en/latest/examples/draw-shapes.html with "Shape type circle"

![image](https://user-images.githubusercontent.com/490840/73928773-f392b980-48d3-11ea-958f-e50c460cf985.png)

The tool should also allow a user to define the radius (using geodesic measurements), and set the circle to a radius of this size. This would require an additional part to the UI - possibly a combobox with some preset values, that also allows for any +ve float value. 

The tool will be used to select parts of a road network in the same way as #73 - so a similar design should be used, with code reuse where possible. See https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/button/DigitizeButtonController.js

PR #177, opened 2020-02-19
Introduces circleSelectionToolbar

jansule:
- adds a button type `Circle` for selecting features by drawing a circle
- adds toolbar that allows the manipulation of the circle radius
- value of radius field always reflects the current unit of the map projection

**Note:** ol 4.6.5 seems to compute the diameter instead the radius on the circle geometry. Therefore the setters and getters for radius have to be divided/multiplied by 2.

solves #171 

Screenshots:

image

image

### DIFF
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -259,10 +259,10 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
         var me = this;
         // deactivate the creation of another circle
         me.drawInteraction.setActive(false);
-        me.circleToolbar = Ext.create('CpsiMapview.view.toolbar.CircleSelection', {
+        me.circleToolbar = Ext.create('CpsiMapview.view.toolbar.CircleSelectionToolbar', {
             feature: evt.feature,
         });
-        me.circleToolbar.on({
+        me.circleToolbar.getController().on({
             circleSelectApply: me.onCircleSelectApply,
             circleSelectCancel: me.onCircleSelectCancel,
             scope: me
@@ -355,7 +355,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
             };
         }
         else if (type === 'Circle') {
-            // ol circle objects consist of a center corrdinate and a radius in the
+            // ol circle objects consist of a center coordinate and a radius in the
             // unit of the projection. In order to convert it into a geoJSON, we have
             // to convert the circle to a polygon first.
             var circleAsPolygon = new ol.geom.Polygon.fromCircle(feat.getGeometry());
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -46,6 +46,18 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
     modifyInteraction: null,
 
     /**
+     * CircleToolbar that will be set
+     * when pressing a button of type `Circle`
+     */
+    circleToolbar: null,
+
+    /**
+     * Parent to add the circleToolbar to. MUST
+     * implement the method `addDocked()`.
+     */
+    circleToolbarParent: null,
+
+    /**
      * The index of the currently active group
      * Only used when `useContextMenu` is true
      */
@@ -62,13 +74,18 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
      * @param {Ext.button.Button} btn The button that has been pressed
      * @param {boolean} pressed The toggle state of the button
      */
-    onToggle: function(btn, pressed) {
+    onToggle: function (btn, pressed) {
         var me = this;
         var view = me.getView();
 
         // guess the map if not given
         if (!me.map) {
             me.map = BasiGX.util.Map.getMapComponent().map;
+        }
+
+        // use default parent for circle toolbar if not defined
+        if (!me.circleToolbarParent) {
+            me.circleToolbarParent = view.findParentByType('cmv_map');
         }
 
         // create a temporary draw layer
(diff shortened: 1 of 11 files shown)


app/controller/button/DigitizeButtonController.js
```diff
@@ -275,6 +354,17 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
                 geometry3857: Ext.JSON.decode(geoJson).geometry
             };
         }
+        else if (type === 'Circle') {
+            // ol circle objects consist of a center corrdinate and a radius in the
```
chrismayer: Typo -> `coordinate`

app/controller/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,90 @@
+/**
+ * This class is the controller for the CircleSelectionController.
+ */
+Ext.define('CpsiMapview.controller.toolbar.CircleSelectionController', {
```
chrismayer: The class name should reflect the the JS file incl. its path:

Due to the path of the source file the class should be named the following: `CpsiMapview.controller.toolbar.CircleSelectionTool`

app/controller/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Handles the click on the `Apply` button by firing
+     * the `circleSelectApply` event of the button that created
+     * the toolbar
+     */
+    handleApply: function () {
+        var me = this;
+        var feat = me.getView().feature;
+        me.getView().fireEvent('circleSelectApply', feat);
```
chrismayer: Why is this `fireEvent` delegated to the view instance?

app/controller/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    /**
+     * Handles the click on the `Cancel` button by firing
+     * the `circleSelectCancel` event of the button that created
+     * the toolbar
+     */
+    handleCancel: function () {
+        var me = this;
+        me.getView().fireEvent('circleSelectCancel');
```
chrismayer: See above

app/controller/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,90 @@
...
+                radius = geom.getRadius() / 2;
+            }
+        }
+        return radius;
+    },
+
+    /**
+     * Handles changes in the numberField by applying the value
+     * as radius for the ol circle feature
+     * @param {Ext numberField} numberField component that fired the event
```
chrismayer:
```suggestion
     * @param {Ext.form.field.Number} numberField component that fired the event
```

app/controller/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Sets the initial value for the `radius` data field in the ViewModel
+     * to the current radius of the ol circle feature
+     */
+    init: function () {
+        var me = this;
+        // geom.setRadius actually is the diameter
+        // therefore we have to multiply by 2
+        var radius = me.getCurrentRadius() * 2;
```
chrismayer: So this var should better be named to `diameter`, or?

app/model/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * This class is the view model for the CircleSelectionToolbar.
+ */
+Ext.define('CpsiMapview.model.toolbar.CircleSelection', {
```
chrismayer: The class name should reflect the the JS file incl. its path:

Due to the path of the source file the class should be named the following: `CpsiMapview.model.toolbar.CircleSelectionTool`

app/view/toolbar/CircleSelectionTool.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * This class is the circle selection of cpsi mapview application
+ * It can be used e.g. for the zone tool
+ */
+/**
+ * Circle Selection
+ *
+ * @class CpsiMapview.view.toolbar.CircleSelection
+ */
+Ext.define('CpsiMapview.view.toolbar.CircleSelection', {
```
chrismayer: The class name should reflect the the JS file incl. its path:

Due to the path of the source file the class should be named the following: `CpsiMapview.view.toolbar.CircleSelectionTool`

app/controller/toolbar/CircleSelectionTool.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/177/commits/b79cde00e7b4d84cc737d59866c7570cad768ee5

app/controller/button/DigitizeButtonController.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/177/commits/b79cde00e7b4d84cc737d59866c7570cad768ee5

app/controller/toolbar/CircleSelectionTool.js
jansule: I had some problems listening to the event, which was based on my little experience with Ext. But was now solved in https://github.com/compassinformatics/cpsi-mapview/pull/177/commits/b79cde00e7b4d84cc737d59866c7570cad768ee5

app/controller/toolbar/CircleSelectionTool.js
jansule: see above, too

app/controller/toolbar/CircleSelectionTool.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/177/commits/b79cde00e7b4d84cc737d59866c7570cad768ee5

app/model/toolbar/CircleSelectionTool.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/177/commits/b79cde00e7b4d84cc737d59866c7570cad768ee5

app/view/toolbar/CircleSelectionTool.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/177/commits/b79cde00e7b4d84cc737d59866c7570cad768ee5

app/controller/toolbar/CircleSelectionTool.js
jansule: I would rather not do it, as I think it might lead to more confusion in the future, as we are still using the (not properly working) `getRadius()` method and are also displaying the actual radius.

E.g. `setRadius(diameter)` could be quite confusing to see for someone else stumbling upon the code.

chrismayer: Agree

geographika:
@jansule - thanks for this - it works very well.

One question - if we save and retrieve the feature later, what would be the best approach to reload it?

Would it be:

- Get a reference to `CpsiMapview.controller.button.DigitizeButtonController`
- Get the `drawLayer` property
- Add the feature directly to this?

jansule:
> it works very well.

Happy to hear that.

> 
> One question - if we save and retrieve the feature later, what would be the best approach to reload it?
> 
> Would it be:
> 
> * Get a reference to `CpsiMapview.controller.button.DigitizeButtonController`
> * Get the `drawLayer` property
> * Add the feature directly to this?

Generally, this should be the way to go. But to make it properly work, there might be changes in the code needed. The toolbar will only be added on the `drawend` event, which might not be fired when just adding a feature. Therefore, I doubt we are able to trigger the selection.

@geographika Did you have the chance to test this approach, yet?

PR #178, opened 2020-02-20
General mechanism for grouped LayerTree

### LINKED ISSUE #145: Allow Layer Grouping
Currently layers are grouped as BASELAYER_OVERLAY and all other overlays in the `makeLayerStore` function in CpsiMapview.view.LayerTree. 

In projects based on the old mapview the tree was semi-manually configured with code such as below:

```
{
    text: "Network Layers",
    expanded: false,
    children: [
        this.getSingleLayerConfig("NetworkEdges"),
        this.getSingleLayerConfig("NetworkNodes"),
        this.getSingleLayerConfig("MissingEdges"),
        {
            text: "Sub Network Layers",
            expanded: false,
            visible: true,
            children: [
                this.getSingleLayerConfig("NetworkEdgesNew"),
                this.getSingleLayerConfig("NetworkNodesNew")
            ]
        },
        {
            plugins: [{
                ptype: 'gx_layercontainer',
                loader: Ext.create("Compass.gis.tree.CustomLoader", { category: "boundaries" })
            }],
            text: "Boundaries",
            expanded: false
        },        
```
image

Loaders were used to load layers by a category setting in the JSON config into a sub-branch of the tree. 

In practice the layer tree usually had to be fine-tuned most of the time to get a certain layer index order in groups. Creating a tree store and placing layers in groups using their `layerKey` from the config file into a hard-coded tree with children would be the most flexible and simplest approach (rather than trying to autogenerate trees from a more complex config file).

chrismayer:
This adds a general mechanism for a grouped `LayerTree`, which is based on a JSON tree structure (see resources/data/layers/tree.json) which defines folders and leafs for the `LayerTree`. The corresponding layers are referenced by the property `layerkey`, so the current layer creation (and thus the layer JSON) can stay untouched.

Here is an example for a tree.json

```json
{
  "id": "root",
  "children": [
    {
      "id": "overlay-folder",
      "title": "Layers",
      "children": [
        {
          "id": "LIGHT_UNIT_MVT",
          "isLeaf": true
        },
        {
          "id": "ACCIDENTS_WFS",
          "isLeaf": true
        },
        {
          "id": "GAS_WFS",
          "isLeaf": true
        }
      ]
    },
    {
      "id": "base-folder",
      "title": "Base Layers",
      "children": [
        {
          "id": "OSM_BACKGROUND",
          "isLeaf": true
        },
        {
          "id": "GREY_BACKGROUND",
          "isLeaf": true
        }
      ]
    }
  ]
}
```
Closes #145

### DIFF
--- a/app/view/LayerTree.js
+++ b/app/view/LayerTree.js
@@ -6,6 +6,7 @@ Ext.define('CpsiMapview.view.LayerTree', {
     xtype: 'cmv_layertree',
     requires: [
         'BasiGX.util.Map',
+        'BasiGX.util.Layer',
         'GeoExt.data.store.LayersTree',
         'CpsiMapview.plugin.BasicTreeColumnLegends',
         'CpsiMapview.plugin.TreeColumnContextMenu',
@@ -15,13 +16,6 @@ Ext.define('CpsiMapview.view.LayerTree', {
         'CpsiMapview.view.menuitem.LayerGrid',
         'CpsiMapview.plugin.TreeColumnStyleSwitcher'
     ],
-
-    viewModel: {
-        data: {
-            baseLayerGroupText: 'Base Layers',
-            overlayGroupText: 'Layers',
-        }
-    },
 
     // So that instantiation works without errors, might be changed during
     // instantiation of the LayerTree.
@@ -90,9 +84,9 @@ Ext.define('CpsiMapview.view.LayerTree', {
         me.callParent([cfg]);
 
         var mapComp = BasiGX.util.Map.getMapComponent();
-        var map = mapComp && mapComp.getMap();
-        if (map) {
-            me.setStore(me.makeLayerStore(map));
+        me.map = mapComp && mapComp.getMap();
+        if (me.map) {
+            me.setStore(me.makeLayerStore(me.map));
         } else {
             Ext.GlobalEvents.on('cmv-mapready', me.autoConnectToMap);
         }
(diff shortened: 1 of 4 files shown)


app/view/LayerTree.js
```diff
@@ -91,8 +85,9 @@ Ext.define('CpsiMapview.view.LayerTree', {
 
         var mapComp = BasiGX.util.Map.getMapComponent();
         var map = mapComp && mapComp.getMap();
-        if (map) {
-            me.setStore(me.makeLayerStore(map));
+        me.map = map;
```
jansule: Minor: why not set `me.map = mapComp && mapComp.getMap()` directly?

geographika:
@chrismayer - love the approach taken!

app/view/LayerTree.js
chrismayer: Correct, could be set directly. I'll change that

chrismayer:
Thanks @geographika! Concept and refinement of the approach was done together with @jansule. So I don't want to take the applause for me alone :wink:

geographika:
@chrismayer - I've successfully set this up in a child application. 

I have one question relating to the "switchlayer" type. Should the layer key of the root be added? 

When I tried this no layer was displayed. To get it working I added both sublayers e.g. LIGHT_UNIT_SWITCH_LAYER_FAR and LIGHT_UNIT_SWITCH_LAYER_CLOSE (otherwise depending on the scale the map opened the layer would be missing). I didn't need to add LIGHT_UNIT_SWITCH_LAYER at all. Is this the approach that should be taken for switchlayers?

chrismayer:
> When I tried this no layer was displayed. To get it working I added both sublayers ...

Yes, this is currently the way to go. I could not find a smarter solution at the moment. Maybe @JakobMiksch has an idea since he implemented the SwitchLayer type.

chrismayer:
> An additional question - can the leaves accept any properties of the NodeInterface? E.g. is it possible to start a layer with "expanded": "true"?

I remember we did that in some other GeoExt projects but it wasn't working out of the box. Let me check if I can find something regarding this.

JakobMiksch:
@geographika yes, it seems that adding both layers is the only possible solution. The reason is probably somewhere in the rather complex switch layer logic.

geographika:
@JakobMiksch @chrismayer - this is fine, the question was for clarification.

PR #179, opened 2020-02-20
147 windows toolbar

### LINKED ISSUE #147: Windows Toolbar and Minimizable Windows
## Overview

The old system allowed windows to be minimised to a "toolbar" - an area at the bottom of the map. These windows could then be restored.
This concept and code was based on the [ExtJS Desktop Example App](https://examples.sencha.com/extjs/6.7.0/examples/classic/desktop/) (version 4) but is still maintained and available in ExtJS 6.x

image

Windows could then have the following properties set:

```
        maximizable: true,
        minimizable: true,
        collapsible: false,
```

Relevant classes in the old codebase:

- [Compass.pms.ux.AbstractTaskBar](https://github.com/compassinformatics/pmsjs/blob/master/src/app/ux/AbstractTaskBar.js)
- [Compass.pms.ux.TaskButton](https://github.com/compassinformatics/pmsjs/blob/master/src/app/ux/TaskButton.js)
- [Compass.pms.utils.BaseFormWindow](https://github.com/compassinformatics/pmsjs/blob/master/src/app/utils/BaseFormWindow.js)
- [TaskButtonsPanel](https://github.com/compassinformatics/pmsjs/blob/master/src/app/ux/TaskButtonsPanel.js)

In addition a "Minimise All" button was available on the toolbar. 

See also various Ext.ux classes such as `Ext.ux.desktop.TaskBar`
## Design Options

- Create as a separate module? See similar choice required for #142 
- Add as a mixin to Windows rather than a base class?

jansule:
Introduces the management for minimizing windows into a windows toolbar. Therefore, two classes were added:

* `CpsiMapview.view.window.MinimizableWindow` which should be used instead of `Ext.window.Window` if minimization should be enabled
* `CpsiMapview.view.toolbar.MinimizedWindows` which is the toolbar that contains the minimized windows

`MinimizableWindow` expects an optional property `minimizeTo` which should contain the toolbar it should minimize to. If that property is not defined, the first found `MinimizedWindows`-toolbar will be used. If no toolbar was added to the application, the windows will minimize into nowhere.

A custom toolbar could also be used but should implement the same events as `MinimizedWindows`.

**Note:** Minimization is only enabled for windows based on `MinimizableWindow`, therefore windows created by dependencies remain untouched.

The toolbar is only visible if at least one window was minimized, and only minimized windows will be listed in the toolbar.

Closes #147 

cpsi-window-manager


app/controller/toolbar/MinimizedWindows.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * This class is the controller for the MinimizedWindows.
+ */
+Ext.define('CpsiMapview.controller.toolbar.MinimizedWindows', {
+    extend: 'Ext.app.ViewController',
+
+    alias: 'controller.cmv_minimized_windows',
```
chrismayer: A hint that this is related to a toolbar view would be nice to have,s o it is more intuitive.
```suggestion
    alias: 'controller.cmv_minimized_windows_toolbar',
```

app/view/toolbar/MinimizedWindows.js
```diff
@@ -0,0 +1,27 @@
...
+ * This class is the toolbar of cpsi mapview application that contains
+ * minimized windows. After a window was resized, the toolbar will be visible.
+ * If a window was minimized without adding this toolbar in advance, the window
+ * will minimize into nowhere.
+ *
+ * @class CpsiMapview.view.toolbar.MinimizedWindows
+ */
+Ext.define('CpsiMapview.view.toolbar.MinimizedWindows', {
+    extend: 'Ext.toolbar.Toolbar',
+    xtype: 'cmv_minimized_windows',
```
chrismayer: Adding `_toolbar`, see above...

test/spec/view/toolbar/MinimizedWindows.spec.js
```diff
@@ -0,0 +1,12 @@
...
+        it('is defined', function() {
+            expect(CpsiMapview.view.toolbar.MinimizedWindows).not.to.be(undefined);
+        });
+
+        it('can be instantiated', function() {
+            var inst = Ext.create('CpsiMapview.view.toolbar.MinimizedWindows');
+            expect(inst).to.be.a(CpsiMapview.view.toolbar.MinimizedWindows);
+        });
+    });
+});
```
chrismayer: Missing new line at the end  (sorry for being picky :see_no_evil:)

app/controller/toolbar/MinimizedWindows.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/179/commits/0b106a127d97601d29ea31d149448e09874ec88e

app/view/toolbar/MinimizedWindows.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/179/commits/0b106a127d97601d29ea31d149448e09874ec88e

test/spec/view/toolbar/MinimizedWindows.spec.js
jansule: solved in https://github.com/compassinformatics/cpsi-mapview/pull/179/commits/0b106a127d97601d29ea31d149448e09874ec88e

app/controller/window/MinimizableWindow.js
```diff
@@ -0,0 +1,55 @@
...
+        var me = this;
+        var minimizeTo = me.getMinimizeToolbar();
+        if (minimizeTo == null) {
+            Ext.log({
+                msg: 'No cmv_minimized_windows_toolbar found. Window might just disappear.',
+                level: 'warn'
+            });
+        }
+        me.getView().setVisible(false);
+        minimizeTo.fireEvent('addMinimizedWindow', me.getView());
```
geographika: Should probably return the function after the warning? Otherwise minimizeTo is null here and throws an error

geographika:
@jansule - thanks for this - works very well!

I added one minor issue I ran into above. 
I have a couple of further questions:

1. Is it possible to make the toolbar transparent, so the map is still visible behind the buttons?
2. The scalebar and "i" OpenLayers icons shift when the toolbar visibility changes. I presume these could be moved using OL CSS to the top of the map to avoid this?

app/controller/window/MinimizableWindow.js
jansule: Was fixed in https://github.com/compassinformatics/cpsi-mapview/pull/185

jansule:
> 1. Is it possible to make the toolbar transparent, so the map is still visible behind the buttons?

Yes and no. It should not be a problem to make the toolbar transparent. However, this will not make the map visible behind the buttons as the map is set above the toolbar. In order to have a transparent toolbar with the map being visible behind would require to add the toolbar as a control to the map. Right now, I am not sure if it will have any side effects if we put Ext components into a control.

It should also be possible to make a css-hack, but since Ext is responsible for the general layouting, we should avoid this if possible.

Another possibility would be to give the toolbar a different background, so that the separation between the map and the toolbar(s) would have a clearer separation.

> 2. The scalebar and "i" OpenLayers icons shift when the toolbar visibility changes. I presume these could be moved using OL CSS to the top of the map to avoid this?

Yes, this is possible. IMHO it looks a little unusual to have the scalebar on top of the map.

We could also add the windows-toolbar to the top of the map. This would look like this:

image

or this (if we also change the background color of the toolbar):

image


@geographika Please let me know, what you think about the different possibilities.

geographika:
@jansule - in ExtJS4 the toolbar was added within the GeoExt MapPanel component (see images below), and we didn't run into any issues. This may no longer be the case, but it would be my preference. Could we attempt this approach and if it breaks OL or ExtJS we can consider the other options?

toolbar_comp2
toolbar_comp1

jansule:
> in ExtJS4 the toolbar was added within the GeoExt MapPanel component

That's good to know. Could you grant me access to the repository (https://github.com/compassinformatics/pmsjs) so I can take a look at the code base?

> Could we attempt this approach and if it breaks OL or ExtJS we can consider the other options?

If it worked then, I am quite confident that it will work here, too.

geographika:
@jansule - you should get an invite to pmsjs. Apologies to @ahennr - I had to remove your access to free up a seat.

Issue #186, opened 2020-02-27
Vector Tile Layer "Light Unit VT" seems to be broken


JakobMiksch:
This `VT` Layer is not working. In the viewer it does not stop loading.

```
    {
      "layerType": "vt",
      "text": "Light Unit VT",
      "layerKey": "LIGHT_UNIT_VT",
      "url": "https://plmonaghandev.compass.ie/mapserver/?mode=tile&tilemode=gmap&tile={x}+{y}+{z}&layers=LightUnit&map.imagetype=mvt",
      "openLayers": {
        "visibility": false
      },
      "format": "MVT",
      "stylesBaseUrl": "https://plmonaghandev.compass.ie/resources/data/styling/",
      "styles": [ "LightUnit_Type.xml", "LightUnit_Owner.xml" ],
      "stylesForceNumericFilterVals": true
    },
```

Console output:

![image](https://user-images.githubusercontent.com/15704517/75480087-f0ee2600-59a0-11ea-8180-b9a9a98e7975.png)

However, the `VT WMS` Layer, which comes from the same source, does work.

I need the `VT`layer for testing #159

geographika:
@JakobMiksch - I think we can probably remove this layer from the JSON file. To get vector tiles working from MapServer @chrismayer implemented an approach using the WMS protocol - see #120. See "layerKey": "LIGHT_UNIT_MVT" layer.

chrismayer:
Closed via #187

PR #187, opened 2020-02-28
Remove unavailable VT layer

chrismayer:
This removes the Vector Tile layer `LIGHT_UNIT_VT` since the service backend is not available anymore. The layer is usable via VT WMS protocol, see layer `LIGHT_UNIT_MVT` (as suggested by @geographika in #186).


chrismayer:
Thanks @geographika.

PR #189, opened 2020-03-03
Display baselayers with radio buttons

### LINKED ISSUE #143: Use Radio Buttons for Base Layers
Currently base layers use check boxes even though only one can be active at a time. Radio buttons would make more sense here.

image

jansule:
BaseLayers will now be displayed with radio buttons instead of checkboxes. Uses property `isBaseLayer` for distinguishing baseLayers from overlays.

image

fixes #143

### DIFF
--- a/app/view/LayerTree.js
+++ b/app/view/LayerTree.js
@@ -160,11 +160,8 @@ Ext.define('CpsiMapview.view.LayerTree', {
 
             me.getRootNode().cascade(function (node) {
                 var data = node.getData();
-                if (data.leaf) {
-                    var isBaseLayer =  data.get('isBaseLayer');
-                    if (isBaseLayer) {
-                        node.addCls('cpsi-tree-node-baselayer');
-                    }
+                if (data.leaf && data.get('isBaseLayer')) {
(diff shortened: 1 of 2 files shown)


app/view/LayerTree.js
```diff
@@ -158,6 +158,16 @@ Ext.define('CpsiMapview.view.LayerTree', {
             });
             me.setStore(groupedLayerTreeStore);
 
+            me.getRootNode().cascade(function (node) {
+                var data = node.getData();
+                if (data.leaf) {
+                    var isBaseLayer =  data.get('isBaseLayer');
+                    if (isBaseLayer) {
```
chrismayer:
```suggestion
                    if (data.leaf && data.get('isBaseLayer')) {
```
surely not blocking

chrismayer: What I wanted to say is to merge the 2 if statements to one.

jansule: done in https://github.com/compassinformatics/cpsi-mapview/pull/189/commits/f7daaa60823018db884ce2a4059da2b4903d2fd2

geographika:
Thanks @jansule !

PR #191, opened 2020-03-03
Add title property

### LINKED ISSUE #159: Allow a title property for style configurations
Currently the [style property ](https://github.com/compassinformatics/cpsi-mapview/blob/c792b0739e56198f18e9f6c7fd5c2f1dab6a0f85/resources/data/layers/default.json#L133)in the layer config file uses the WMS style names, and these are displayed as radio buttons. 

image
```
          "styles": [
            "Unit Type",
            "Owner"
          ],
```
Ideally a title could be (optionally?) set for these configurations to override the text and provide a more human friendly name. E.g. 

```
          "styles": [
            "Unit Type",
            {"name": "Owner", "title" "Ownership Body"}
          ],
```

JakobMiksch:
fixes #159 
- checks if a style in the config file has a provided title, if yes it will be applied to the layer tree
- but it is still allowed to provide no title

### DIFF
--- a/app/view/layer/StyleSwitcherRadioGroup.js
+++ b/app/view/layer/StyleSwitcherRadioGroup.js
@@ -40,17 +40,34 @@ Ext.define('CpsiMapview.view.layer.StyleSwitcherRadioGroup', {
 
         // create a 'radiofield' for each style connected to the layer
         Ext.each(layerStyles, function (layerStyle) {
+
+            var layerTitle;
+            var layerName;
+
+            // check if style has title property
+            if(layerStyle['title'] && layerStyle['name']){
+                // has title property
+                layerTitle = layerStyle['title'];
+                layerName = layerStyle['name'];
+            } else{
+                // does not have title property
+                // title generated from stlye name
+                layerTitle = me.getLayerStyleLabel(layerStyle);
+                layerName = layerStyle;
+            }
+
             radioButtons.push({
                 name: 'sldstyle' + salt,
-                boxLabel: me.getLayerStyleLabel(layerStyle),
-                inputValue: layerStyle,
-                checked: me.getCheckedState(layerStyle),
+                boxLabel: layerTitle,
+                inputValue: layerName,
+                checked: me.getCheckedState(layerName),
                 listeners: {
                     change: me.onStyleChange,
                     scope: me
                 }
             });
-        });
+        }
+        );
 
         me.items = radioButtons;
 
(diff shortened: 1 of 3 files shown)


chrismayer:
I just checked out the changes of this PR and ran the `test` script with no issues. SO it seems like an issue with Travis.

jansule:
> I just checked out the changes of this PR and ran the `test` script with no issues. SO it seems like an issue with Travis.

It might be a caching problem, if it works on your local setup. So clearing the cache on travis and rerunning the job might solve the problem.

chrismayer:
> So clearing the cache on travis and rerunning the job might solve the problem.

:tada: This solved it. Thanks @jansule.

@JakobMiksch Would you be available to resolve the merge conflict? Then this is good to go IMHO. TIA!

JakobMiksch:
@chrismayer I resolved the conflict. It was basically to integrate the title property stuff into the config file

chrismayer:
Thank you, I'll merge now

geographika:
@JakobMiksch - thanks for this. I've run into one issue. When I click the layer visibility checkbox (rather than directly selecting a style) an object is passed via the querystring e.g. 
`&STYLES=%5Bobject%20Object%5D&`

This then throws an issue in MapServer:

`msWMSLoadGetMapParams(): WMS server error. Style ([object Object]) not defined on layer.`

Screencast at: https://drive.google.com/file/d/1rpiSHQINvobss4T7Y8lZqTUgLXzbnVH0/view

@chrismayer  how does a default style get set for a layer? Some seem to have a radio button selected by default whereas others do not. There doesn't seem to be anything in default.json to set this.

geographika:
Note this also causes an issue when right-clicking a layer with LayerLabels:

```js
        if (wmsParams && !Ext.isEmpty(wmsParams.STYLES) &&
            wmsParams.STYLES.indexOf(me.labelClassName) !== -1) {
```

```
LayerLabels.js?_dc=1584139971923:67 Uncaught TypeError: wmsParams.STYLES.indexOf is not a function

at constructor.onAfterrender (LayerLabels.js?_dc=1584139971923:67)
````
 - as STYLES is an object rather than string.

chrismayer:
@geographika The two issue you reported in here should be fixed by #207. Could you please check if it works in your projects?

geographika:
@chrismayer - with this and #207 both issues have been addressed when testing locally. Thanks!

PR #192, opened 2020-03-03
Add default values to config file

### LINKED ISSUE #172: Allow default settings to be applied to the JSON config
The default.json config file used to generate the layer list can often contain many duplicated settings. 
It should be possible to apply default settings to layers to avoid these being duplicated. 

A hierarchy of defaults would be useful e.g. apply to all layers, then apply to all layers of "layerType": "wms". Defaults could be stored in additional .json files or within a derived application. 

```json
    {
      "layerType": "wms",
      "helpPage": "OSM",
      "qtip": "An OSM based WMS layer",
      "iconCls": "map",
      "isBaseLayer": false,
      "isDefaultBaseLayer": false,
      "text": "A WMS",
      "url": "https://ows.terrestris.de/osm-gray/service",
      "serverOptions": {
        "layers": "OSM-WMS"
      },
      "openLayers": {
        "maxResolution": 1222.99245234375,
        "numZoomLevels": 12,
        "opacity": 0.7,
        "projection": "EPSG:900913",
        "visibility": false,
        "zoomOffset": 7
      }
    },
```

JakobMiksch:
fixes #172 
- the config file is enhanced by sections for default values. 
- a general section applies default values for all types of layers
- each layertype has an own section which specific defaults
- all defaults values will be overwritten by the values of each single layer.


resources/data/layers/default.json
```diff
@@ -274,4 +295,4 @@
       ]
     }
   ]
-}
+}
```
jansule: Very minor: Please add newline

JakobMiksch: Alright. Fixed in this commit 35ac69f2919081e3d922723b870c225e9c5a6afe

jansule: Yes, thank you. Feel free to merge.

geographika:
@JakobMiksch - thanks for this. I have one question/comment - if I have nested objects in the defaults for example the openLayers object below:

```json
  "defaults": {
    "wms": {
      "serverOptions": {
        "version": "1.3.0"
      },
      "openLayers": {
        "visibility": false
      }
    },
```

Then this setting is ignored for any layers which already have the openLayers nested object. For example the layer below will have `openLayers.visibility === true`  (so ignoring the default above, even though there is no explicit visibility setting for the layer):

```json
    {
      "layerType": "wms",
      "text": "Borehole WMS",
      "layerKey": "BOREHOLE_WMS",
      "serverOptions": {
        "layers": "Boreholes",
        "version": "1.3.0"
      },
      "openLayers": {
        "singleTile": false,
      },
```

Just wondering if [Ext.Object.merge](https://docs.sencha.com/extjs/6.2.0/modern/Ext.Object.html#method-merge) might allow the recursive merging of objects rather than using `Ext.apply` (came across this in #200).

JakobMiksch:
@geographika I replaced the `Ext.apply` with `Ext.Object.merge` and tested it. It seems to fix the problem. Thanks for the hint!
I pushed my commit to this branch https://github.com/JakobMiksch/cpsi-mapview/commit/9b98298da912d86fdf9fd47c3d4fa51cad70adec because I was not sure how to add my commit to this closed pull request. Otherwise I can open a new pull request. Just let me know.

geographika:
@JakobMiksch - a new pull request is probably easiest. Thanks!

JakobMiksch:
here it is: https://github.com/compassinformatics/cpsi-mapview/pull/206

Issue #182, opened 2020-02-21
Use GeoStyler libs via npm 


chrismayer:
The GeoStyler libs used in here (geostyler-openlayers-parser & geostyler-sld-parser) are currently manually created browser builds and thus not updateable easily. In the meantime the GeoStyler and its related projects (e.g. parsers) are available at [npm](https://www.npmjs.com/search?q=keywords:geostyler), so we should use them via npm to have an automatic update strategy for the GeoStyler libs used.

TODOs:

- Add corresponding entries to `dependencies` in package.json
- Reference the browser build files from `node_modules` in the `js` section in the `app.json`

geographika:
Closed with #193

PR #193, opened 2020-03-03
Add geostyler dependencies via node

geographika:
Partially addresses #182 - will also need to update app.json


geographika:
> 
> 
> Did you check if the specified versions work properly? The `geostyler-openlayers-parser` has a peerDependency on Openlayers 6 (see https://github.com/geostyler/geostyler-openlayers-parser/blob/v2.0.1/package.json#L33-L35) while we are on 4.6.5.

Thanks for pointing that out. The last version to support 4.6.5 appears to be [v0.21.0](https://github.com/geostyler/geostyler-openlayers-parser/blob/v0.21.0/package.json)
I'll update this.

PR #194, opened 2020-03-04
144 Vector Layer Labels MenuItem

### LINKED ISSUE #144: Vector Layer Labels MenuItem
Currently CpsiMapview.view.menuitem.LayerLabels only supports WMS layers. 
The same class would ideally also work for vector based layers. 

image

Note any updates would also need to be taken into account the following in CpsiMapview.view.layer.StyleSwitcherRadioGroup

if (layer.get('labelsActive') === true) {
    newStyle += ',' + labelClassName;
}

jansule:
Adds handling for displaying labels of vector data (i.e. layerTypes `wfs`, `vt`, `vtwms`).

Adds following new properties to `default.json`:
* `sldUrlLabel` *`String`*: Url to Style with labeled version of `sldUrl`
* `styles[x].label` *`String*`: Name of the labeled version of `styles[x].name`

Prioritization was taken over from `Layer.js`. Thereby, `styles` will be preferred if it and all of its required parameters are defined.

@geographika 
fixes #144 

---
**Update:** With the upcoming merge of https://github.com/compassinformatics/cpsi-mapview/pull/191 the `styles` property will change anyway. So I will harmonize the structure after https://github.com/compassinformatics/cpsi-mapview/pull/191 was merged.

**Update2:** harmonized structure with #191. Updated PR description accordingly.

@chrismayer open for review


geographika:
@jansule - is this ready to test or commit, or should I wait for a review first?

app/view/layer/StyleSwitcherRadioGroup.js
```diff
@@ -151,6 +151,32 @@ Ext.define('CpsiMapview.view.layer.StyleSwitcherRadioGroup', {
             } else if (layer.get('isWfs') || layer.get('isVt')) {
 
                 var sldUrl = layer.get('stylesBaseUrl') + newStyle;
+
+                /**
+                 * Helper function to check if at least one style
+                 * has a non-empty label property.
+                 * @param {Array<{label: String}>} styles Styles object from layerConfig to check
```
chrismayer:
```suggestion
                 * @param {Object[]} styles Styles object from layerConfig to check
```

app/view/menuitem/LayerLabels.js
```diff
@@ -124,5 +262,73 @@ Ext.define('CpsiMapview.view.menuitem.LayerLabels', {
...
+            return;
+        }
+
+        return stylesBaseUrl + labelStyle.label;
+    },
+
+    /**
+     * Helper function to check if at least one style
+     * has a non-empty label property.
+     * @param {Array<{label: String}>} styles Styles object from layerConfig to check
```
chrismayer:
```suggestion
     * @param {Object[]} styles Styles object from layerConfig to check
```

app/view/layer/StyleSwitcherRadioGroup.js
```diff
@@ -151,6 +151,32 @@ Ext.define('CpsiMapview.view.layer.StyleSwitcherRadioGroup', {
...
+                 */
+                var hasLabels = function (styles) {
```
chrismayer: Maybe re-using the `hasLabels` function of `CpsiMapview.view.menuitem.LayerLabels`? Or adding a util function to `CpsiMapview.util.Legend`?

chrismayer:
I expect mutual side-effects due to #201 and #207

jansule:
> I tried to make this changes work in the appliaction, unfortunately without success. The GAS WFS remains without a label on the map, when activating via context menu:
> 
> Am I missing something?

@chrismayer The `label` checkbox is activated on a layer, if at least one of the referenced styles also provides a label style. In your case, only the `CUSTOM TITLE` style provides a labelled style, while the `Gas` style does not.

If you activate the `CUSTOM TITLE` style and then enable labeling, you should be able to see the effect.

I guess it might be better to check if the `label` option should be shown or hidden for each style separately.

chrismayer:
OK, then I figured out an edge case. What I did:

1. Set GAS layer to visible
2. Activate the labels via context menu
3. Switch style to "CUSTOM TITLE"

Then no labels are shown. If switch point 2. and 3. it works as expected and labels are shown.

jansule:
> One quick question - I presume that the SLD with the label also has to include all feature styling as in the demo layer?
> 
> ```
>           "name": "Test_Alternative_Gas_Style.xml",
>           "title": "CUSTOM TITLE",
>           "label": "Test_Alternative_Gas_Style_label.xml"
> ```
> 
> So Test_Alternative_Gas_Style_label.xml contains everything in Test_Alternative_Gas_Style.xml plus an additional TextSymbolizer?

@geographika exactly. OpenLayers only allows a single StyleFunction for a Style (which will be used for e.g. attributive styling). Therefore, we have to provide a separate SLD that (re-)defines the style and adds the TextSymbolizers.

geographika:
@jansule - thanks for the clarification on the labels. 
Is this ready to merge?

Just a note to self - the legend appears broken for the WFS layer as the mapview project expects any WFS layer to have an associated WMS layer of the same name on the server (which is the case when using MapServer, possibly not GeoServer). This was to avoid having to create vector based legends which are not part of GeoExt3. 

The GAS layer does not have a GAS style:

https://ows.terrestris.de/geoserver/osm/wfs?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetLegendGraphic&FORMAT=image%2Fpng&TRANSPARENT=TRUE&SLD_VERSION=1.1.0&LAYER=osm:osm-fuel&STYLE=Gas

```xml
<ServiceExceptionReport version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://ows.terrestris.de/geoserver/schemas/wms/1.3.0/exceptions_1_3_0.xsd">
<ServiceException>
java.lang.NullPointerException: request.getStyle() request.getStyle()
</ServiceException>
</ServiceExceptionReport>
```

So I think this is ready to merge?
I can look at setting up a new demo test layer following the merge with vector label functionality.

jansule:
> So I think this is ready to merge?

@geographika Currently, the handling for the case when not all styles of a layer also provide a label style is not optimal (see https://github.com/compassinformatics/cpsi-mapview/pull/194#pullrequestreview-380180251).

If the current state is fine for you (for now), I can resolve the merge conflicts and we can merge. Otherwise, there would still be a little work to do.

geographika:
@jansule - would the idea be to merge two different SLD files on the client-side using JS?

As MapServer is the backend for all our projects at the moment I'm seeing if we can fix this on the server-side. 

Currently, there is (limited) support for creating SLD using a WMS service in MapServer - however all styles are returned. Label/TextSymbolizer support is not currently working, but will be fixed in the next MapServer release. 

For example https://w08-mapserver.compass.ie/mapserver/?map=/MapServer/apps/mapview-demo/example.map&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetStyles&LAYERS=Boreholes
returns 4 UserSyles in a single request. 

I'm looking at allowing only certain styles to be returned using a STLYES= parameter. See [this post](https://lists.osgeo.org/pipermail/mapserver-dev/2020-March/016124.html). 

If the tool could allow a "thematic" UserStyle plus  a "label" UsersStyle to be applied when labels are turned on, would this be the best long-term option?

In the short-term if I manually create a SLD file with multiple UserStyles with one named `<Name>labels</Name>` could we implement this approach?

Let me know if you need clarifications on the above.

jansule:
@geographika This would be the cleanest approach in the sense that we would reduce style repetitions to a minimum. However, we would also lose flexibility, as there would be only a single label style defined for all styles of a layer. Depending on the used colors, etc. it might be necessary to define custom labels for each style to be readable (or visually appealing).

I would rather recommend to keep the concept as is, and add the limitation that all styles of a label MUST specify a label style in order to enable labeling.

If wanted, a utility function on the server could do the merging and persist the merged style. Thereby, we would still have the flexibility of changing the label style of a layer, while reducing the efforts to create all of them by hand.

Also, @chrismayer might have some useful insights?

chrismayer:
I also would favour to leave the style definition by the server (by SLD). This gives us flexibility to have different styles per layer and we decouple the style generation from our code, which is good IMHO. So a change in the style would not force us to change the code and thus re-deploy.
So providing separate "Label SLDs" that (re-)define the styles are a but of double work, but I gues it is worth doing so.

geographika:
@chrismayer @jansule - I'm fine with duplicating the styles and make the server do the work and duplicate all styles with a label as in the current implementation. 

Are multiple UserStyles currently applied from a single SLD file? If this is the case then this will fit perfectly with updating MapServer to allow a request including `STYLES=Gas_Style,Gas_Style_Label`

This would return the `Gas_Style `and the `Gas_Style_Label `in a single SLD document, but as two UserStyles. 

Currently this is not implemented in MapServer so I am scripting the generation of SLDs with Python so this can be looked at in a future phase if required.

jansule:
@geographika currently, we are using a single SLD file for each style.

Working with your approach should be possible, but would require additional work on the client. I would suggest to at first follow the two-files approach, so that we have a proper handling for labeling vector layers. 

In a second step, we could then adopt the single-file approach and provide a separate PR for that.

geographika:
> 
> 
> @geographika currently, we are using a single SLD file for each style.
> 
> Working with your approach should be possible, but would require additional work on the client. I would suggest to at first follow the two-files approach, so that we have a proper handling for labeling vector layers.
> 
> In a second step, we could then adopt the single-file approach and provide a separate PR for that.

@jansule this approach sounds good to me. Please merge when you get a chance.

jansule:
@geographika 

I just solved the merge conflicts and changed the code to only enable labeling if ALL styles provide  a label style.

This should be ready to merge now, if you don't see any conflicts.

Issue #183, opened 2020-02-23
WFS Layer "LightUnit" seems to be broken


JakobMiksch:
This WFS Layer does not display in CPSI-Mapview. In QGIS or with OGR it does not open either. The WMS from the same URL however works for me.

```
    {
      "layerType": "wfs",
      "text": "Light Unit WFS",
      "layerKey": "LIGHT_UNIT_WFS",
      "url": "https://plmonaghandev.compass.ie/mapserver/?",
      "geometryProperty": "msGeometry",
      "featureType": "LightUnit",
      "sldUrl": "resources/data/styling/LightUnit_Unit_Type.xml",
      "openLayers": {
        "visibility": false
      }
    },
```

geographika:
@JakobMiksch - we're in the process of setting up stable services just for this project in #173
I've tried to fix that layer in the meantime. In theory any valid WFS could be used for testing if that still fails.

JakobMiksch:
I used a different WMS for testing. Thanks. You can close this issue if you want.

geographika:
Closed with #201

Issue #173, opened 2020-02-12
Create standalone WxS Services for Demo


geographika:
Currently several layers in the mapview demo project are broken due to database changes. 
A new set of services using flat files will be created specifically for the project to avoid breaking changes. These will be setup with CORS to avoid requiring a proxy. 

Broken layers:

* Light Unit WFS
* Accidents

geographika:
Closed with #201

PR #201, opened 2020-03-13
Use new layers from demo project

geographika:
A new standalone project has been created at https://github.com/compassinformatics/mapview-demo and running at http://w08-mapserver.compass.ie/mapserver/?map=\MapServer\apps\mapview-demo\example.map& to serve layers for the cpsi-mapview project. 

These demo layers will not be subject to changes, unlike several WMS and WFS services currently used in the project. 

Broken layers have been removed and replaced with new layers, while retaining the functionality in the demo. 

Should address #183 and #173

### DIFF
--- a/resources/data/layers/default.json
+++ b/resources/data/layers/default.json
@@ -153,50 +153,6 @@
     },
     {
       "layerType": "switchlayer",
-      "layerKey": "LIGHT_UNIT_SWITCH_LAYER",
-      "visibility": false,
-      "layers": [
-        {
-          "layerType": "wms",
-          "text": "Light Unit Switch Layer (far)",
-          "layerKey": "LIGHT_UNIT_SWITCH_LAYER_FAR",
-          "serverOptions": {
-            "layers": "LightUnit",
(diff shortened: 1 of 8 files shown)


geographika:
@chrismayer @JakobMiksch @jansule - any objections to merging this?
The previous layers were tied to a specific project which will be subject to breaking changes. These new demo layers are used just for this project so will be more stable (and new layer types can be set up as required). CORS is also enabled on the demo server.

resources/data/layers/default.json
```diff
@@ -145,64 +153,57 @@
...
-        "version": "1.1.1"
-      },
-      "openLayers": {
-        "singleTile": true,
-        "visibility": false
-      }
-    },
     {
       "layerType": "wfs",
       "text": "Test time WFS (needs proxy)",
```
chrismayer: Do we need this anymore?

geographika: @chrismayer - thanks for the feedback. I've added in the vtwms layer type and removed the time layer as suggested.

PR #205, opened 2020-03-17
Make default values work recursivly 

JakobMiksch:
This fix also applies the properties of the underlying objects to the layer configuration

see #172


JakobMiksch:
actually added the same changes to https://github.com/JakobMiksch/cpsi-mapview/commit/9b98298da912d86fdf9fd47c3d4fa51cad70adec
closing again

PR #207, opened 2020-03-24
Fix detection of initial style parameters for layers

chrismayer:
This PR fixes the detection of the initial style parameter for WMS, WFS and VT layers.
    
The detection now checks if the first entry in the style config list is an object (with STYLES / SLD file name and UI alias) or if the STYLES / SLD file name is directly provided.

The edge case was introduced by #191.


app/factory/Layer.js
```diff
@@ -216,8 +216,13 @@ Ext.define('CpsiMapview.factory.Layer', {
         var styles = layerConf.serverOptions.styles;
         var activatedStyle;
         if (Ext.isArray(layerConf.styles) && layerConf.styles.length) {
-            styles = layerConf.styles[0];
-            activatedStyle = layerConf.styles[0];
+            // check if first possible SLD style list is an object (with STYLES
+            // name and UI alias) or if the STYLES name is directly provided.
+            var firstStyle = layerConf.styles[0];
+            var stylesWmsParam =
+                Ext.isObject(firstStyle) ? firstStyle.name : firstStyle[0];
```
JakobMiksch: One observation: here you call `firstStyle[0]` whereas in the other edits (WFS and VT) you only call `firstStyle`. Just wondering if this intentionally

chrismayer: Good catch, I fixed and rebased...thanks

chrismayer:
@JakobMiksch -I just address your findings and resolved the conflicts. Are you willing to have another quick look if this is good to go now?

JakobMiksch:
I check it out and tested it. All layers seem to work well except of `Country WFS` and `Gas WFS`. First I thought, it is a problem of the WFS, but the the WFS inside the switch layer works. So, it might be another reason.

geographika:
> 
> 
> I check it out and tested it. All layers seem to work well except of `Country WFS` and `Gas WFS`. First I thought, it is a problem of the WFS, but the the WFS inside the switch layer works. So, it might be another reason.

@JakobMiksch @chrismayer - I think this is a layer config issue as noted at https://github.com/compassinformatics/cpsi-mapview/pull/194#issuecomment-604376221 so I'm happy to merge this.

chrismayer:
Thanks you for your input guys! I'll merge now.

PR #210, opened 2020-03-24
Fix broken WMS legend image for labelled layers

### LINKED ISSUE #209: Legend for labelled WMS layers not shown
When activating labels for a WMS layer via context menu and then changing the style the legend image breaks:

grafik

chrismayer:
Fixes the broken WMS legend image for labelled layers.

Closes #209

### DIFF
--- a/app/util/Legend.js
+++ b/app/util/Legend.js
@@ -52,6 +52,11 @@ Ext.define('CpsiMapview.util.Legend', {
                 return;
             }
 
+            // Remove possible duplicates created by adding labels
+            // (CpsiMapview.view.menuitem.LayerLabels) because GetLegendGraphic
+            // accepts only one layer in its LAYER param
+            layers = LegendUtil.getUniqueLayersParam(layers);
+
             requestParams += 'SERVICE=WMS&VERSION=1.3.0&REQUEST=GetLegendGraphic&';
             requestParams += 'FORMAT=image%2Fpng&TRANSPARENT=TRUE&SLD_VERSION=1.1.0&';
             requestParams += 'LAYER=' + layers + '&';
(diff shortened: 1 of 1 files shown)


app/util/Legend.js
```diff
@@ -119,6 +124,20 @@ Ext.define('CpsiMapview.util.Legend', {
...
+     * @param  {String} layers WMS LAYERS parameter value, e.g. (layer1,layer2)
+     * @return {String}        WMS LAYERS parameter without duplicates
+     */
+    getUniqueLayersParam: function (layers) {
+        var layerList = Ext.isArray(layers) ? layers : layers.split(',');
+        // remove any duplicate layer names
+        var reducedLayerList = Ext.Array.unique(layerList);
+
+        return reducedLayerList.join(',');
+    },
```
geographika: Extra comma here should be removed?

chrismayer: I don't think so. `join(',')` handles this and avoids a trailing comma. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join). Or do I oversee something?

chrismayer:
Thanks for the review @JakobMiksch! Waiting for @geographika 's go (because of the open question) and the I am going to merge this.

app/util/Legend.js
geographika: Sorry I meant line 141. This used to break things on IE7 :)

chrismayer: Oh now I see ... yes the good old IE 7 times :laughing: I guess the JS builder takes care of this, but to be save I'll remove this.

chrismayer:
What is failing here is the Azure pipeline, not the unit tests. I assume the Azure pipeline should not be executed here at all, correct @geographika?.

geographika:
@chrismayer - the pipeline is used to build a minified version. It was working fin eon my branch before merging. Please ignore while I update it and go ahead with merging this.

chrismayer:
Thanks for clarification @geographika. I'll merge now.

PR #211, opened 2020-03-25
Set minimizedWindows transparent & position it over map

jansule:
This is a follow up of #179 

The minimizedWindows toolbar is now transparent and was positioned over the map.

image

### DIFF
--- a/app/view/main/Map.js
+++ b/app/view/main/Map.js
@@ -28,9 +28,6 @@ Ext.define('CpsiMapview.view.main.Map', {
     }, {
         xtype: 'cmv_mapfooter',
         dock: 'bottom'
-    }, {
-        xtype: 'cmv_minimized_windows_toolbar',
-        dock: 'bottom'
     }],
 
     items: [{
@@ -56,6 +53,8 @@ Ext.define('CpsiMapview.view.main.Map', {
         listeners: {
             afterrender: 'afterMapRender'
         }
+    }, {
+        xtype: 'cmv_minimized_windows_toolbar'
     }],
 
     /**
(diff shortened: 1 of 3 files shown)


jansule:
@geographika does this match your expectation?

geographika:
Thanks for this @jansule. 
There is one issue at the moment - the minimised toolbar buttons are placed on top of each other when there are multiple windows. 

I tried this with the grid and the Streetview tool  (to activate this tool you need a Google API key and update app.json), although it appears any windows will minimise to the same location. 

image

jansule:
> There is one issue at the moment - the minimised toolbar buttons are placed on top of each other when there are multiple windows.

Thanks for the info. I will take a look at that.

jansule:
@geographika with the latest changes, the toolbar looks like this. As Ext is now handling the horizontal layouting of the toolbar, this should work regardless of the number of minimized windows.

image

geographika:
@jansule - just tested this and it works very well. I'm happy to merge/close this. Thanks!

Just to confirm I'm not missing anything - if I'm adding this functionality to a project based on cpsi-mapview then I need to ensure the following:

1. Include `xtype: 'cmv_minimized_windows_toolbar'` within a `gx_map` component
2. Update the `Main.scss` with the .cmv_minimized_windows* styles
3. Ensure all windows in the application inherit from `CpsiMapview.view.window.MinimizableWindow`

jansule:
@geographika yes, that is correct. Happy to hear that you are satisfied with the solution.

chrismayer:
I just checked this out and tested. Works as expected. Thank you @jansule! Please merge at will...

Issue #241, opened 2020-07-06
Layer Tree not configured in MS Edge


geographika:
This issue only appears to occur in older versions of MS Edge - recreated in `Microsoft Edge 44.19041.1.0`
This does not seem to happen in `Version 83.0.478.58 (Official build) (64-bit)`  or Chrome, IE, or Firefox. 

The issue manifests itself as a flat layer tree - no folders, or nesting defined in the tree.json are taken into account. 

The cause of this is that the layers are loaded into the map when default.json is loaded [initComponent](https://github.com/compassinformatics/cpsi-mapview/blob/375e72c043d34c168d225cd06514f60c665cbc2d/app/view/main/Map.js#L198)

In Edge it appears that this event occurs after the tree.json is loaded and applied. This means when the [layer tree](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/LayerTree.js#L264) is searching for layers using:

```
var mapLyr = BasiGX.util.Layer.getLayerBy('layerKey', child.id);
```

No layers are found (they do not yet exist in the map), and no tree structure is created.

geographika:
Fixed with #242 - but a review would be useful.

PR #242, opened 2020-07-06
Apply tree structure after all layers are loaded. See #241

geographika:
This pull request waits until the map layers have been added to the map before applying the tree structure. 

This allows the structure to work correctly in older version of the Edge browser. It does however rely on the `cmv-init-layersadded` event being called and so does not allow a local configuration. 

In addition this pull request renames "Root" to "root" as this name is used [here in the TreeColumnStyleSwitcher](https://github.com/compassinformatics/cpsi-mapview/blob/7cb03dbf484582e10f9ee5cd718e56b2862c8bbe/app/plugin/TreeColumnStyleSwitcher.js#L145) and is case-sensitive.

### DIFF
--- a/app/view/LayerTree.js
+++ b/app/view/LayerTree.js
@@ -98,7 +98,7 @@ Ext.define('CpsiMapview.view.LayerTree', {
 
     /**
      * Guesses the mapcomponent and assigns the appropriate layers store, if one
-     * could be guessed. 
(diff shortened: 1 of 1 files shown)


geographika:
Have also run into the same issue (race condition) in Chrome. Adding this fix for now.

PR #266, opened 2020-08-25
new field type CpsiMapview.view.field.YearField

leonardospina:
Custom field to display years in a YYYY format. Displays text aligned to right.


app/view/field/YearField.js
```diff
@@ -0,0 +1,16 @@
+Ext.define('CpsiMapview.view.field.YearField', {
+    extend: 'Ext.form.field.Number',
+    xtype: 'cmv_yearfield',
+    fieldStyle: 'text-align: right;', // TODO has no effect?
```
geographika, line 4: Can you check if this is still a valid ExtJS property? If not then it should be removed

app/view/field/YearField.js
```diff
@@ -0,0 +1,16 @@
...
+    // textAlign: 'right', // property not available
```
geographika, line 5: Remove commented code

geographika (changes requested):

geographika:
Thanks @leonardospina - could you add a brief description of the change as part of the pull request?

app/view/field/YearField.js
leonardospina: It is!

app/view/field/YearField.js
leonardospina: Commented code removed

geographika (approved):

PR #281, opened 2020-09-30
Linear reference

### LINKED ISSUE #249: Linear Referencing Display
The required functionality is as follows:

- a single OL vector feature is in a layer tied to a grid
- the grid has a store where each record (row in the grid) has a startMeasure and endMeasure
- selecting a row on the grid highlights that section of the line feature e.g. if a record has startMeasure 405 and endMeasure 670 then the section of the line between 405m and 670m is highlighted

image

image

Creating measures along a line seem to be available in turf.js - see http://turfjs.org/docs/#along and https://openlayers.org/en/latest/examples/turf.html
OL also has a [getCoordinateAtM](https://openlayers.org/en/latest/apidoc/module-ol_geom_LineString-LineString.html#getCoordinateAtM) function which might be enough?

Features will:

- never be multipart linestrings
- have a single start and end (no T shapes for example)

A grid mixin or similar would be ideal for integrating into larger systems.

simonseyock:
~~I added the functionality as a plugin. This PR contains 2 commits, the second should not get merged as it contains an example usage. I remove the second commit when this is ok to merge.~~ (see below)

This resolves #249 

LineSegmentHighlightPlugin

methods:
highlightSlice - highlights a slice in the given geometry
removeHighlight - removes the highlight
setStyle - set the style of the highlight


simonseyock:
I removed the second commit and moved it to another branch linear-reference-example. So this one is save to merge if reviewed.

geographika:
@simonseyock - I think it would be good to add in the demo to this pull request and the main project. `cpsi-mapview` is a dependency for real-world projects rather than an application itself, so when deployed on its own it is a "kitchen-sink" app to display all available functionality. 

A button on the UI to open the `CpsiMapview.view.linearReference.LinearReference` grid in a `CpsiMapview.view.window.MinimizableWindow` would be perfect.

geographika:
@simonseyock - a question you may be able to provide a link/advice on. 
In the application that requires this I will be getting several LineString geometries from a GeoJSON response. 
Does OpenLayers have methods to join several connected LineString together into a single LineString object - not a MultiLineString but a valid LineString object. The end coordinate of one LineString will be the start of the next. 
If not I guess I can handle this on the server, but there are 2 sources of data so I'd prefer to code it once on the client. 

https://github.com/compassinformatics/cpsi-mapview/blob/8ce377bef202754579abadde27ab615e3bbfbd9b/app/view/linearReference/LinearReference.js#L46

simonseyock:
There is no native method, but it can be done buy just joining the coordinates:
```
    // assuming lines is an array of LinesString geometries
    var coordinates = lines.reduce(function (coordinates, nextLine) {
        return coordinates.concat(nextLine.getCoordinates().slice(1));
    }, lines[0].getCordinates()[0]);
    var line = new ol.geom.LineString(coordinates);
```

simonseyock:
I added in an example in a window.

geographika (approved):

app/plugin/LineSliceHighlight.js
```diff
@@ -0,0 +1,83 @@
...
+                // end is in this segment
+                coordinates.push(segment.getCoordinateAt((end - length) / segmentLength));
+            }
+
+            length += segmentLength;
+        });
+        return new ol.geom.LineString(coordinates);
+    },
+
+    highlightSlice: function (geometry, start, end) {
```
chrismayer, line 52: Could you please add API docs with a short description? Also for `removeHighlight` and `setStyle`.

app/view/lineSliceGridExample/LineSliceGrid.js
```diff
@@ -0,0 +1,113 @@
...
+            dataIndex: 'endMeasure',
+            renderer: meterRenderer
+        },
+        {
+            text: 'Text',
+            dataIndex: 'text'
+        }
+    ],
+    listeners: {
+        itemclick: function (_, record) {
```
chrismayer, line 46: Wouldn't it make sense that the plugin registers the event itself, so that by using the plugin the functionality is automatically available in the father grid?

app/view/main/Main.js
```diff
@@ -44,5 +45,7 @@ Ext.define('CpsiMapview.view.main.Main', {
         }
     }, {
         xtype: 'cmv_header'
+    }, {
+        xtype: 'cmv_line_slice_grid_window'
```
chrismayer, line 49: How about instantiating the `cmv_line_slice_grid_window` in the handler of `app/view/lineSliceGridExample/LineSliceGridButton.js` instead of hanging it in the main view?

chrismayer (commented):
Very nice @simonseyock! I added some questions we might want to briefly discuss before merging.

app/view/lineSliceGridExample/LineSliceGrid.js
simonseyock: Yes that would be possible, but would only work if the properties in the used model are named exactly `startMeasure` and `endMeasure`. We could test if that is the case and if someone would want to use other properties he or she could still call the `highlightSlice` method manually.

simonseyock: Also the plugin does not have access to the geometry as it is. At the moment the method `highlightSlice` accepts a geometry as a parameter. Instead we could have a `setGeometry` method - then this could be done.

chrismayer: OK understand, just leave it as it is. Even more flexible the way it is done now.

chrismayer (approved):

PR #282, opened 2020-09-30
Refactored LayerTree to LayerTreeController

simonseyock:
This PR moves code from the LayerTree view into a new LayerTreeController as it belongs from a MVC point of view into a controller.

It also extracts code from `updateLayerNodeUi` into a new function `getNodeForLayer`.

### DIFF
--- a/app/view/LayerTree.js
+++ b/app/view/LayerTree.js
@@ -5,8 +5,6 @@ Ext.define('CpsiMapview.view.LayerTree', {
     extend: 'Ext.tree.Panel',
     xtype: 'cmv_layertree',
     requires: [
-        'BasiGX.util.Map',
-        'BasiGX.util.Layer',
         'GeoExt.data.store.LayersTree',
         'CpsiMapview.plugin.BasicTreeColumnLegends',
         'CpsiMapview.plugin.TreeColumnContextMenu',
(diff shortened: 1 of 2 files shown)


app/view/LayerTree.js
```diff
@@ -54,228 +55,33 @@ Ext.define('CpsiMapview.view.LayerTree', {
...
     },
 
     /**
-     * Updates the layer node UI for the given layer.
+     * Gets the layer node UI for the given layer.
      *
      * @param  {ol.layer.Base} layer The layer to update in the tree
      */
-    updateLayerNodeUi: function (layer) {
+    getNodeForLayer: function (layer) {
```
geographika, line 72: Is there any reason not to also move this (and updateLayerNodeUi) into the controller?

geographika (commented):
Looks good to me. Just a question on why `updateLayerNodeUi` and `getNodeForLayer` aren't also moved (there may be a good reason)?

app/view/LayerTree.js
simonseyock: I think the `getNodeForLayer` is a view specific thing because it gets the view element (the node) corresponding to the layer. It is also called from a different component and it this way I was able to find it with `Ext.ComponentQuery`.

geographika: Thanks - makes sense.

geographika (approved):

app/view/LayerTree.js
chrismayer, line 72: This somehow refactors and adds a new method, correct? Just for a better understanding since the PR description indicates that this PR is just moving existing methods to a new controller.

simonseyock: yes the method is new. But the functionality was extracted from `updateLayerNodeUi` and `updateLayerNodeUi` now uses the `getNodeForLayer` method.

chrismayer: Okay, thanks for clarification. Not always easy in the diff view :wink:

chrismayer (approved):

chrismayer:
Good to go, thanks @simonseyock! Will merge now.

Issue #274, opened 2020-09-22
Performance Issues with Layer Tree Legends


geographika:
When a layer's visibility changes the UI hangs briefly, the layer tree disappears and is then redrawn. When there are lots of layers this can take 1s or more to complete, and delays other UI actions (e.g. a form being hidden). 

## Possible Cause

The initial trigger is any part of the application changing the layer visibility - in GeoExt the LayerTreeNode - onLayerVisibleChange is called. 

In the mapview project `getLegendHtml` is called for every layer in the tree.

https://github.com/compassinformatics/cpsi-mapview/blob/1d3687e678e0ba62ea87d05ca051244dce3a4b02/app/plugin/BasicTreeColumnLegends.js#L48

The following always appear to be `undefined` and so `LegendUtil.createGetLegendGraphicUrl(layer)` is called (which I believe is taking the time).

```
var legendUrl = layer.get('legendUrl');
var w = layer.get('legendWidth');
var h = layer.get('legendHeight');
```

Maybe these should be set on the first calculation, so they are cached?
However there are several other plugins used in the LayerTree that may be the cause. 

https://github.com/compassinformatics/cpsi-mapview/blob/1d3687e678e0ba62ea87d05ca051244dce3a4b02/app/view/LayerTree.js#L37

chrismayer:
With regard to the [comment](https://github.com/compassinformatics/cpsi-mapview/pull/284#issuecomment-701591173) of @geographika:

Yes it seems that the complete tree is re-rendered in case a property (mostly visibility of a layer) gets changed. Due to my reaserch it is even worse since the `getLegendHtml` function of the `CpsiMapview.plugin.BasicTreeColumnLegends` is called several (~6) times in case one layer is set visible / invisible. Maybe it is related to cascading down the whole tree structure due to `toggleParentNodes` as suggested in your comment. I can dig deeper into this, but I am not sure if we can get around this.

geographika:
@chrismayer - if you can find out why the full tree is redrawn that would help to decide what to do. On testing a normal treepanel it should not need to refresh all nodes when the backing store is updated. 
The full refresh also causes #287

chrismayer:
Did some more investigations on this:

[...]

PR #284, opened 2020-09-30
Cache legend images

chrismayer:
This caches the legend images as dataURLs in an internal lookup, so the images only have to be queried at the first activation of the legend graphic in the tree.

@geographika Could you please give it a try in an application with a large amount of layers in order to test if this brings a performance boost for the layer tree? TIA! 

Relates to #274

### DIFF
--- a/app/util/Html.js
+++ b/app/util/Html.js
@@ -1,0 +1,28 @@
+/**
+ * Static util class for HTML related operations.
+ */
+Ext.define('CpsiMapview.util.Html', {
+
+    statics: {
+        /**
+         * Add events to HTML elements in a cross browser compatible manor.
+         *
+         * Credits: http://stackoverflow.com/a/21140463
+         *
+         * @param {DOMElement} htmlElement The element to attach the event to.
+         * @param {String} eventName The name of the event to attach.
+         * @param {Function} eventFunction The handler for the `eventName` on
+         *     the passed `htmlElement`.
+         */
+        addEvent: function (htmlElement, eventName, eventFunction) {
+            if (htmlElement.addEventListener) { // Modern
+                htmlElement.addEventListener(eventName, eventFunction, false);
+            } else if (htmlElement.attachEvent) { // Internet Explorer
+                htmlElement.attachEvent('on' + eventName, eventFunction);
+            } else { // others
+                htmlElement['on' + eventName] = eventFunction;
+            }
+        }
+    }
+
+});
--- a/app/util/Legend.js
+++ b/app/util/Legend.js
@@ -181,7 +181,7 @@ Ext.define('CpsiMapview.util.Legend', {
      */
     cacheLegendImgAsDataUrl: function(legendUrl, layerKey) {
         var legendImage = new Image();
-        legendImage.crossOrigin="anonymous";
+        legendImage.crossOrigin = 'anonymous';
         CpsiMapview.util.Html.addEvent(
             legendImage, 'load', function() {
 
--- a/app/util/Legend.js
+++ b/app/util/Legend.js
@@ -6,7 +6,8 @@
 Ext.define('CpsiMapview.util.Legend', {
     alternateClassName: 'LegendUtil',
     requires: [
-        'BasiGX.util.Object'
+        'BasiGX.util.Object',
+        'CpsiMapview.util.Html'
     ],
 
     singleton: true,
--- a/app/util/Legend.js
+++ b/app/util/Legend.js
@@ -150,5 +150,63 @@ Ext.define('CpsiMapview.util.Legend', {
         var reducedLayerList = Ext.Array.unique(layerList);
 
         return reducedLayerList.join(',');
+    },
+
+    /**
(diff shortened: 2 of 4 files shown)


geographika:
@chrismayer - is there a new file missing `CpsiMapview.util.Html`? Getting xtemplate errors at the moment.

chrismayer:
Sorry you are right...added missing file with e56eb63

geographika:
@chrismayer - it does seem to refresh a little quicker - and cuts down on lots of `GetLegendGraphic` requests. 
The cache is a nice idea. There may be a requirement for "data-driven" legends or scale dependent legends in the future, but those could probably be excluded with a config setting. 

In regard to overall performance it appears the entire tree is redrawn whenever any layer changes visibility - any idea if this is required?
Could this full redraw be caused by https://github.com/geoext/geoext3/pull/62 - `toggleParentNodes` seems to loop through every layer setting properties on each model:

```
                    parent.set('__toggleMode', 'ol3'); // prevents recursion
                    parent.set('checked', true);
                    parent.set('__toggleMode', 'classic');
```

Could it be this causing the full redraw?
I recently had an issue with models having properties set to `dirty` when not required which can be avoided with an options object and `dirty` property - see [set docs](https://docs.sencha.com/extjs/6.7.0/classic/Ext.data.Model.html#method-set). 
However, this may be required to keep children/parent checkboxes in synch. 

```
parent.set('__toggleMode', 'classic', {dirty: false});
```

I tried with the alternative [ol3 toggle mode](https://geoext.github.io/geoext3/master/docs/#!/api/GeoExt.data.store.LayersTree-cfg-folderToggleMode) but it still seems to redraw all nodes on the tree.

app/util/Html.js
```diff
@@ -0,0 +1,28 @@
...
+         * Add events to HTML elements in a cross browser compatible manor.
+         *
+         * Credits: http://stackoverflow.com/a/21140463
+         *
+         * @param {DOMElement} htmlElement The element to attach the event to.
+         * @param {String} eventName The name of the event to attach.
+         * @param {Function} eventFunction The handler for the `eventName` on
+         *     the passed `htmlElement`.
+         */
+        addEvent: function (htmlElement, eventName, eventFunction) {
```
geographika, line 17: I presume ExtJS doesn't already have a function for this that could be used?

app/util/Legend.js
```diff
@@ -150,5 +151,63 @@ Ext.define('CpsiMapview.util.Legend', {
...
+                canvas.height = height;
+
+                var ctx = canvas.getContext('2d');
+                ctx.drawImage(this, 0, 0);
+                var legendDataUrl = canvas.toDataURL('image/png');
+
+                // store in cache for later reuse
+                CpsiMapview.view.LayerTree.legendImgLookup[layerKey] = legendDataUrl;
+
+                LegendUtil['legendLoading_' + layerKey] = false;
```
geographika, line 205: Note to self: this is where the isLoading flag is set to false

geographika (approved):

app/view/LayerTree.js
```diff
@@ -18,6 +18,10 @@ Ext.define('CpsiMapview.view.LayerTree', {
 
     controller: 'cmv_layertree',
 
+    statics: {
```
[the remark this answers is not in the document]
chrismayer: Yes, this was the idea and the reference it quite easier. But I can check if I bind it to the LayerTree instance, if you prefer.

app/util/Html.js
chrismayer: As far as I know not, but I'll double check.

chrismayer:
Thanks for the feedback @geographika. I'll suggest to switch the general LayerTree performance discussion to #274 since this here is only only aspect regarding to optimize performance.

app/view/LayerTree.js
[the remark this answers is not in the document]
geographika: No need for any changes it was just for my understanding

app/util/Html.js
geographika: Guess if there was it would be part of https://docs.sencha.com/extjs/6.2.0/classic/Ext.dom.Element.html ?

geographika:
@chrismayer - is this ready to merge? It stops lots of unnecessary requests so helps speed up the application. 
The function has access to the layer so if there were scale-dependent legends it looks like it should be simple enough to disable caching with an option.

app/util/Html.js
chrismayer: Probably yes. I had a quick look and it seems `Ext.Dom.Element` has the normal `on` event binding method. Maybe it is worth giving this a try so we could get rid of the HTML util class.

chrismayer:
> @chrismayer - is this ready to merge?

Has some potential for optimization, like getting rid of the HTML util class (see comment above) or prevent more `GetLegend` requests at startup. But this can surely be done in follow-up PRs in a later stage. As already stated this should bring another performance boost, so I pledge to merge. Or what do you think @geographika ?

geographika:
@chrismayer - agree!

PR #285, opened 2020-09-30
Feature Attribute Grouping Plugin

simonseyock:
This is implemented reacting to the hover event and has only a very basic styling option (color).

This is maybe a good point to get some feedback @geographika

Configuration is done in the `default.json` in the layer object like this:

```
"grouping": {
    "attribute": "some-attribute",
    "color": "green"
}
```

Starting this grouping on hovering has two disadvantages, though:
* the highlight is removed when the mouse is moved or if implemented differently as soon as another feature is hovered.
* It can interfere with the get feature info (the picture below)

What styling options are needed apart from color?

It is to mention that this grouping would out of the box not work with clustered layers.

Peek 2020-09-30 15-20
grouping+gfi

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -118,6 +118,8 @@ Ext.define('CpsiMapview.factory.Layer', {
             mapLayer.set('opacitySlider', allowOpacitySlider);
             // the xtype of any associated grid
             mapLayer.set('gridXType', layerConf.gridXType);
+            // attribute grouping config
+            mapLayer.set('grouping', layerConf.grouping);
         }
 
         return mapLayer;
--- a/app/plugin/FeatureAttributeGrouping.js
+++ b/app/plugin/FeatureAttributeGrouping.js
@@ -3,51 +3,116 @@ Ext.define('CpsiMapview.plugin.FeatureAttributeGrouping', {
     alias: 'plugin.cmv_feature_attribute_grouping',
     pluginId: 'cmv_feature_attribute_grouping',
 
+    /**
+     * can have the values 'move' or 'click'
+     */
+    startGroupingEvent: 'move',
+
+    /**
+     * can have the values 'move', 'click' or 'context'
+     * @property {string}
+     */
+    endGroupingEvent: 'move',
+
     initGrouping: function (mapCmp, olMap) {
         var me = this;
 
-        var groupingLayer = new ol.layer.Vector({
+        this.mapCmp = mapCmp;
+        this.olMap = olMap;
+
+        this.layer = new ol.layer.Vector({
             source: new ol.source.Vector(),
             map: olMap
         });
 
-        function clearSource() {
-            groupingLayer.getSource().clear();
+        var group = function (evt) {
+            me.startGrouping(evt);
+        };
+        var ungroup = function () {
+            me.endGrouping();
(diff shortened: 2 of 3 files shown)


geographika:
@simonseyock - not had a chance to test locally yet, but the approach looks very nice. 
A couple of initial questions:

- Is it possible to keep both tooltips and feature highlighting?
- Styling options that would be useful are outline color (although that seems to be aleady in the demo GIF above) and outline width. 
- Does the plugin take into account all geometry types including polys and lines?

> the highlight is removed when the mouse is moved or if implemented differently as soon as another feature is hovered.

Is this something that could be configurable at the application level or layer level? The only issue with clearing on a new feature would be how to clear without making a new selection. Any ideas on how this could be handled?

simonseyock:
> Is it possible to keep both tooltips and feature highlighting?

Yes. Just the feature tooltip might at times hide some highlights.

> Styling options that would be useful are outline color (although that seems to be aleady in the demo GIF above) and outline width.

I add them

> Does the plugin take into account all geometry types including polys and lines?

Not yet. If other geometries should be highlighted as well the styling might get more complex. For now I would just add strokes in the chosen color. We can see what we need from there.

simonseyock:
> > Does the plugin take into account all geometry types including polys and lines?
>
> Is this something that could be configurable at the application level or layer level? The only issue with clearing on a new feature would be how to clear without making a new selection. Any ideas on how this could be handled?

Yes that could be configurable in the component or in the layer. I would prefer configuration in the component so it does behave consitently.

For clearing i can think of 2 options:
1. provide a context menu on right click
2. clear on any click (left or right) on the map and on pressing the ESC key. Maybe also clearing on leaving the map.

simonseyock:
the configuration can be done like this now:
```
"grouping": {
    "attribute": "some-attribute",
    "strokeColor": "red",
    "strokeWidth": 2
}
```

I added two new properties to the component: `startGroupingEvent` and `endGroupingEvent`. Both can be set to `'click'` and `'move'` and the `endGroupingEvent` can additionally be `'context'` in which case a context menu is opened on right click, if a grouping is active.

simonseyock:
these properties can be set to the plugin like this:
```
    plugins: [
        {
            ptype: 'cmv_feature_attribute_grouping',
            startGroupingEvent: 'click',
            endGroupingEvent: 'context'
        }
    ],
```

geographika:
@simonseyock - very nice!
I've successfully integrated with a full application (see image below). I guess I have to watch out for any other map tools that use context menus and use a different GroupingEvent method?


image

geographika (approved):

simonseyock:
>  guess I have to watch out for any other map tools that use context menus and use a different GroupingEvent method?

As it is written right now, this is true. Alternatively a central approach to render the context menu on the map could be thought of. This could then compose a context menu for all interaction options at this point. This might be useful if there will be more map interactions in the future that need a context menu.

My idea would be a component where you can register right click listeners that get the mapEvent and return a menu item if one should be shown.

geographika:
> My idea would be a component where you can register right click listeners that get the mapEvent and return a menu item if one should be shown.

Sounds a good approach and something to look at in a future sprint. 
I'm happy to merge this as-is for now.

PR #293, opened 2020-10-02
Apply default values for tree nodes

### LINKED ISSUE #291: Apply defaults to tree.json Configs
Relating to #289 - it would be useful to apply defaults to all items in tree.json, in the same way as #206 was added to default.json. 
There is no concept of layer type in the tree.json structure, so it would just be a single `defaults > general` section. 

```
{
  "defaults": {
    "general": {
      "iconCls": "fa-bars"
    },
```

chrismayer:
This allows to define a section `"defaults"` > `"general"` having default values for tree node configurations, so duplication of properties can be avoided.

CAUTION: The general structure of tree.json is changed with this PR and needs adaption in existing applications:

before:
```json
{
  "id": "root",
  "children": [
    {
      "id": "overlay-folder",
...
```
after:
```json
{
  "defaults": {
    "general": {}
  },
  "treeConfig": {
    "id": "root",
    "children": [
      {
        "id": "overlay-folder",
...
```
Resolves #291

### DIFF
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -27,6 +27,15 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
      */
     initLayersAdded: false,
 
+    /**
+     * Holds the default values for tree nodes of the config file tree.json.
+     * Will be set in #makeLayerStore function.
+     *
+     * @private
+     * @readonly
+     */
+    treeConfDefaults: {},
+
     constructor: function () {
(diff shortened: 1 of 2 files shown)


geographika (approved):

geographika:
@chrismayer - looks good to me. Feel free to merge - I'll be updating app configs this morning to the new structure.

chrismayer:
Thanks for reviewing @geographika !

PR #294, opened 2020-10-02
Add missing requires and remove baselayer class code

geographika:
@chrismayer - a couple of fixes required when integrating with larger application. 

1. Add missing requires. 
2. `isBaseLayer` code removed as can now be set in tree.json (old approach no longer works):

```
          {
            "id": "OSM_BACKGROUND",
            "text": "OpenStreetMap",
            "iconCls": "fa-map",
            "cls": "cpsi-tree-node-baselayer"
          },
```


chrismayer (approved):
LGTM, thank you!

PR #295, opened 2020-10-02
Add cls 'cpsi-tree-node-baselayer' to base layer tree configs

chrismayer:
Adds `"cls": "cpsi-tree-node-baselayer"` to the tree configurations of the BaseLayers, so the application looks as before (follow-up for #294).

### DIFF
--- a/resources/data/layers/tree.json
+++ b/resources/data/layers/tree.json
@@ -114,14 +114,14 @@
             "leaf": true,
             "text": "OpenStreetMap",
             "qtip": "An OpenStreetMap based background layer",
-            "iconCls": "map"
+            "cls": "cpsi-tree-node-baselayer"
           },
           {
             "id": "GREY_BACKGROUND",
             "leaf": true,
             "text": "Grey Background",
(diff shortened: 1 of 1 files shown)


resources/data/layers/tree.json
```diff
@@ -114,14 +114,14 @@
             "leaf": true,
             "text": "OpenStreetMap",
             "qtip": "An OpenStreetMap based background layer",
-            "iconCls": "map"
+            "cls": "cpsi-tree-node-baselayer"
```
geographika, line 117: Both cls and iconCls can be kept on the nodes - the CSS sets the checkbox style, whereas the iconClass adds an additional icon

chrismayer: I know, but we don't have any meaningful icons (besides the default ones) in the demo application. Setting `"iconCls": "map"` would end in no visible icon.

geographika: ok sorry was looking at my other application! Feel free to merge

PR #299, opened 2020-10-05
Allow the Add WMS Window to be configured

geographika:
This pull request has a few changes following integration into a large application. 

1. Makes the config for the Add WMS window cofigurable on the LayerTree view. As this is where the xtypes are set I also moved the requirements for the default `cmv_minimizable_window` and `cmv_add_wms_form` here. 
2. Adds a missing controller requires (JS errors otherwise) - `CpsiMapview.controller.addWms.AddWmsFormController`

Applications using the mapview library can then fully configure the Add WMS window from the LayerTree:

```js
    addWmsWindowConfig: {
        xtype: 'cmv_minimizable_window',
        title: 'Add External Map Layer',
        items: [{
            xtype: 'cmv_add_wms_form',
            width: 400,
            wmsBaseUrls: [
                'http://gis.epa.ie/geoserver/wms',
                'https://ows.terrestris.de/osm/service',
            ]
        }]
    }
```

3. Makes the form hidden rather than destroyed when closing. If destroyed then a new folder is added to the layer tree each time the form is reopened and a layer added as `this.layerGroup` is null each time a new `CpsiMapview.controller.addWms.AddWmsFormController` is created.

### DIFF
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -291,13 +291,24 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
         var me = this;
 
         if (pressed) {
-            this.addWmsWindow = Ext.create(me.getView().addWmsWindowConfig);
+            if (!this.addWmsWindow) {
+                this.addWmsWindow = Ext.create(me.getView().addWmsWindowConfig);
+                this.addWmsWindow.on('hide', function () {
+                    button.setPressed(false);
+                });
+            }
             this.addWmsWindow.show();
-            this.addWmsWindow.on('close', function () {
-                button.setPressed(false);
-                me.addWmsWindow = null;
-            });
+
         } else if (this.addWmsWindow) {
+            this.addWmsWindow.hide();
+        }
+    },
+
+   /**
+   * Destroy any associated windows when this component gets destroyed
+   */
+    onBeforeDestroy: function () {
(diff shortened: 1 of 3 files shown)


app/controller/LayerTreeController.js
```diff
@@ -293,19 +291,24 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
...
+                });
+            }
             this.addWmsWindow.show();
-            this.addWmsWindow.on('close', function () {
-                button.setPressed(false);
-                me.addWmsWindow = null;
-            });
+
         } else if (this.addWmsWindow) {
+            this.addWmsWindow.hide();
```
simonseyock, line 303: For this to work the `closeAction` needs always to be `'hide'`. Maybe it should be fixed to that?

simonseyock:
I was thinking about creating a mini-component `ToggleWindowButton` that would get a window config and does the toggling. It could also be used for the example grid for the linear referencing.
Apart from the one minor issue this looks good to me.

app/controller/LayerTreeController.js
geographika: Good point - changed back to `close` and will let the closeAction handle that.

geographika:
@simonseyock - thanks for review. Merging now.
Yes a `ToggleWindowButton` sounds a handy util.

PR #307, opened 2020-10-08
Labels on the UI completed with the "parent type" name which the file is attached to

leonardospina:
Messages on the UI look more specific referring to the main object


app/form/field/YearField.js
```diff
@@ -1,7 +1,8 @@
 Ext.define('CpsiMapview.form.field.YearField', {
```
geographika, line 1: @leonardospina - I think there is a problem with your branch - this file keeps getting reverted on each pull request. Probably best to branch again from compassinformatics/master ?

PR #308, opened 2020-10-12
WMS GetFeatureInfoWindow

### LINKED ISSUE #251: WMS GetFeatureInfo Tool
Create a new Information Tool allowing users to click on features on the map and return attribute information. 
This would be similar to old GeoExt functionality such as that outlined in:

https://girona-geoext-workshop.readthedocs.io/en/latest/stores/getfeatureinfo.html

image

Possibly relevant code in the [BasiGX.plugin.Hover](http://terrestris.github.io/BasiGX/master/docs/#!/api/BasiGX.plugin.Hover-method-getFeatureInfoCount) tool. 

For WMS layers the GetFeatureInfo protocol will be used to return data in GeoJSON format. Example URLs will be added to this issue. 

To discuss:

- Expand top layer by default?
- Combine with vector features in the same location? Leave for future phase.

simonseyock:
Resolves #251

This is still missing some documentation but can already be tested. It can be added as a plugin to the Map component:
```
    plugins: [
        {
            ptype: 'cmv_feature_info_window'
        }
    ],
```

Only layers will be queried, that have the property `featureInfoWindow` set to `true` in the `default.json`.

It looks like this:

Peek 2020-10-12 15-10

### DIFF
--- a/app/plugin/FeatureInfoWindow.js
+++ b/app/plugin/FeatureInfoWindow.js
@@ -1,0 +1,142 @@
+Ext.define('CpsiMapview.plugin.FeatureInfoWindow', {
+    extend: 'Ext.plugin.Abstract',
+    alias: 'plugin.cmv_feature_info_window',
+    pluginId: 'cmv_feature_info_window',
+
+    requires: [
+        'Ext.window.Window',
+        'BasiGX.util.Map'
+    ],
+
+    init: function () {
+        var me = this;
+
+        var mapComp = BasiGX.util.Map.getMapComponent();
+        var map = mapComp.getMap();
+
+        map.on('singleclick', function (evt) {
+            var layers = [];
+            var resolution = map.getView().getResolution();
+            var projection = map.getView().getProjection();
+            map.forEachLayerAtPixel(evt.pixel, function (layer) {
+                layers.push(layer);
+            }, undefined, function (layer) {
+                return layer.get('featureInfoWindow');
+            });
+
+            if (layers.length) {
+                var window = me.showFeatureInfo(map, evt);
+
+                layers.forEach(function (layer) {
+                    var url = layer.getSource().getGetFeatureInfoUrl(evt.coordinate, resolution, projection, {
+                        INFO_FORMAT: 'geojson'
+                    });
+
+                    Ext.Ajax.request({
+                        url: url
+                    }).then(function (response) {
+                        var featureCollection = JSON.parse(response.responseText);
+                        window.add(me.createFeatureCollectionPanel(featureCollection));
+                    });
+                });
+            } else {
+                if (me.highlightSource) {
+                    me.highlightSource.clear();
+                }
+                if (me.window) {
+                    me.window.destroy();
+                }
+            }
+        });
+    },
+
+    showFeatureInfo: function (map, evt) {
+        var me = this;
+
+        if (!this.highlightSource) {
+            this.highlightSource = new ol.source.Vector();
+            new ol.layer.Vector({
+                source: this.highlightSource,
+                style: new ol.style.Style({
+                    image: new ol.style.Circle({
+                        radius: 5,
+                        stroke: new ol.style.Stroke({
+                            color: 'red',
+                            width: 2
+                        })
+                    })
+                }),
+                map: map
+            });
+        } else {
+            this.highlightSource.clear();
+        }
+
+        var feature = new ol.Feature(new ol.geom.Point(evt.coordinate));
+        this.highlightSource.addFeature(feature);
+
+        if (this.window) {
+            this.window.destroy();
+        }
+
+        this.window = Ext.create('Ext.window.Window', {
+            layout: {
+                type: 'accordion',
+                titleCollapse: false,
+                animate: true
+            },
+            width: 300
+        });
+        this.window.show();
+
+        this.window.on('close', function () {
+            me.highlightSource.clear();
+        });
+
+        return this.window;
+    },
+
+    createFeatureCollectionPanel: function (featureCollection) {
+        var me = this;
+
+        return Ext.create('Ext.panel.Panel', {
+            title: featureCollection.name,
+            items: featureCollection.features.map(function (geojsonFeature) {
+                return me.createFeatureGrid(geojsonFeature);
+            })
+        });
+    },
+
+    createFeatureGrid: function (geoJsonFeature) {
+        var props = geoJsonFeature.properties;
+
+        var data = Object.keys(props)
+            .filter(function (key) {
+                return key !== 'geometry';
+            })
+            .map(function (key) {
+                return {
+                    key: key,
+                    value: props[key]
+                };
+            });
+
+        var store = Ext.create('Ext.data.Store', {
+            data: data
+        });
+
+        return Ext.create('Ext.grid.Panel', {
+            store: store,
+            columns: [
+                {
+                    dataIndex: 'key'
+                },
+                {
+                    dataIndex: 'value'
+                }
+            ],
+            scrollable: true,
+            layout: 'fit'
+        });
+    }
+});
--- a/app/plugin/FeatureInfoWindow.js
+++ b/app/plugin/FeatureInfoWindow.js
@@ -38,12 +38,12 @@ Ext.define('CpsiMapview.plugin.FeatureInfoWindow', {
             return layer.get('featureInfoWindow');
         });
 
-        var window = me.openFeatureInfoWindow(map, evt);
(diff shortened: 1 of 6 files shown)


simonseyock:
We can have either a fixed height which will result in a less moving view, but has the disadvantage of maybe being to small if many layers return feature infos. Or we could use a flexible layout, the height of the window will change depending on the size of the grid. I will see if I find other layout options.

geographika:
Very nice!

I have this working in the larger application. A few comments/feedback:

- Should use the layer title from the layer tree rather than the WMS name
- Expand the column name to fit all available space (same for value). Field name to take priority (use "flex" property?)
- Allow a user to select and copy and paste a value from the grid
- Possibly have contained in a CpsiMapview.view.window.MinimizableWindow? If minimised window should display again on a map click. However it would then make sense to display a placeholder such as "No results found" rather than hiding the window when a map click returns no results
- when there is one layer it can't be collapsed - maybe the collapse button should be hidden in this case?
- as per the Travis CI there are 2 "plugins" configs in the Map.js file. There should probably also be a [null check here](https://github.com/compassinformatics/cpsi-mapview/blob/bd24a05b185014bae00dcaf082281c0b6d9b3cd3/app/view/main/Map.js#L304) if `cmv_feature_attribute_grouping` is not used (or maybe a `Ext.log` comment). 
- what is the best approach for working with other tools? For example the network solver tool when active adds points but also pops-up the WMS results. Same for the measure tools. 

An enhancement for a future sprint:

- Allow via the JSON config to link a result to a form, and open it using the ID.

app/plugin/FeatureInfoWindow.js
```diff
@@ -0,0 +1,142 @@
...
+                    key: key,
+                    value: props[key]
+                };
+            });
+
+        var store = Ext.create('Ext.data.Store', {
+            data: data
+        });
+
+        return Ext.create('Ext.grid.Panel', {
```
annarieger, line 128: You can possibly better make use of [`BasiGX.view.grid.FeaturePropertyGrid`](https://github.com/terrestris/BasiGX/blob/master/src/view/grid/FeaturePropertyGrid.js), which has some nice features like clickable href links or attribute mapping, instead of creating a `Ext.grid.Panel` here.

simonseyock: Good point.

simonseyock:
With the proposed change of Anna Rieger the grid would be flipped as seen in the following image:

wms feature info

This has the advantage that multiple features at the same point would be shown much nicer.

geographika:
@simonseyock @annarieger - the grid view looks good to me, and if it can take advantage of existing classes even better.

simonseyock:
I noticed I used the wrong component. With the one proposed it would look like this:

wms feature info

What would you prefer? The one I posted earlier has some unneeded functionality (sorting, filtering) and the horizontal layout would be in the most cases not the nicest to read. But for multiple feature infos in the same layer it would be better.

geographika:
@simonseyock - the last one looks the nicest, so I'd go with that.

image

simonseyock:
Filling the remaining space is possible. That the name column takes priority is not possible with standard extjs configuration as I see it. The propertygrid only allows very limitied control over the columns.
There are 2 possibilities as I see it: 
1. Setting a fixed bigger value for the name column. This is the only setting that is originally intended for configuring the column size.
2. Setting a flex value will assign a bigger fixed portion of the available space (i.e. always 50%). I could set this only via private method which should be avoided.

For now I will use 1. and use a bigger value (150px instead of 115px)

simonseyock:
I added the requested changes. For the work with other tools I propose `disableDefaultClick()`, `enableDefaultClick()` and `isDefaultClickEnabled()` methods on the map component.
Then the network solver for example can call `disableDefaultClick()` and the feature info tool calls `isDefaultClickEnabled()` and knows that the window should not get displayed.
This way the network solver does not need to know what other component uses the click event by default.

We could also generalize it to other event names (it could be also useful for `pointermove` or `pointerrest`) by designing the methods like `disableDefaultAction(eventName)`, `enableDefaultAction(eventName)` and `isDefaultActionEnabled(eventName)`.

geographika:
@simonseyock - the `disableDefaultClick()` etc. suggestion sounds good. 

I've been testing the tool - looks very slick - thanks. 

I have run into a few issues with layout errors. The following block seems to cause errors when a second layer is introduced. 

```
        if (layers.length <= 1) {
            window.setLayout('fit');
        } else {
            window.setLayout({
                type: 'accordion',
                titleCollapse: false,
                animate: true
            });
        }
```

The error occurs in the ExtJS framework:

```js
    beginLayout: function(ownerContext) {
        var me = this,
            owner = me.owner,
            smp = owner.stretchMaxPartner,
            style = me.innerCt.dom.style, // error thrown here "TypeError: Cannot read property 'dom' of undefined"
```

`window.getLayout().innerCt` is undefined in some cases.

app/plugin/FeatureInfoWindow.js
```diff
@@ -0,0 +1,136 @@
...
+
+        var format = new ol.format.GeoJSON();
+
+        map.forEachLayerAtPixel(evt.pixel, function (layer) {
+            layers.push(layer);
+        }, undefined, function (layer) {
+            return layer.get('featureInfoWindow');
+        });
+
+        var window = me.openFeatureInfoWindow(map, evt);
```
geographika, line 41: Maybe use win rather than window (the global variable) to avoid confusion when looking at the code in the future?

app/plugin/FeatureInfoWindow.js
```diff
@@ -0,0 +1,136 @@
...
+                    INFO_FORMAT: 'geojson'
+                });
+
+                Ext.Ajax.request({
+                    url: url
+                }).then(function (response) {
+                    var features = format.readFeatures(response.responseText);
+                    window.add(me.createFeaturePanels(mapComp, layer, features));
+                }).then(undefined, function (error) {
+                    console.error(error);
```
geographika, line 69: Use the `Ext.log` function here? Some WMS layers could return XML errors which it would be useful to log.

app/plugin/FeatureInfoWindow.js
simonseyock: Agreed

app/plugin/FeatureInfoWindow.js
simonseyock: I put it in there. I just noticed that ExtJs seems to have the bad habit of not reporting any exceptions raised inside of promises so I put the manual error reporting there.

simonseyock:
The layout problem could be fixed by initializing it with the accordion layout.

The default click enable ended up as a property on the ol-map object (`enableDefaultClick`) as I would had to extend the BasiGX map component otherwise.

app/plugin/FeatureInfoWindow.js
```diff
@@ -0,0 +1,143 @@
...
+                });
+            });
+        }
+    },
+
+    openFeatureInfoWindow: function (map, evt) {
+        var me = this;
+
+        if (!this.highlightSource) {
+            this.highlightSource = new ol.source.Vector();
```
chrismayer, line 82: Is there a special reason why `highlightSource` is a member var since it is just used in `openFeatureInfoWindow`? If yes, please declare and comment it around line 12.

chrismayer (commented):
Thanks for this @simonseyock, looks really good. I added some questions / remarks you might want to have a look at.

app/plugin/FeatureInfoWindow.js
simonseyock: It is stored so it is not recreated every time.

app/plugin/FeatureInfoWindow.js
```diff
@@ -0,0 +1,143 @@
...
+        'Ext.Component',
+        'CpsiMapview.view.window.MinimizableWindow',
+        'BasiGX.view.grid.FeaturePropertyGrid',
+        'BasiGX.util.Map'
+    ],
+
+    init: function () {
+        var me = this;
+
+        var mapComp = BasiGX.util.Map.getMapComponent();
```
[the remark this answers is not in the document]
simonseyock: It is actually used as plugin for `CpsiMapview.view.main.Map` which then contains the map component. But it is also possible to use it as a plugin for the `gx_map`. This might make more sense.

simonseyock:
I resolved the remarks. The plugin is now for the `gx_map` component.

Also I found another problem with the layout which should now be fixed.

geographika:
Tested in a couple of derived applications and working very well - thanks @simonseyock for an excellent feature!
Merging.

PR #309, opened 2020-10-12
Cache legends based on layer plus STYLE

geographika:
Relates to #284 - currently legends are cached per layer, but individual WMS STYLEs can have different legends. This pull request caches at the LAYER/STYLE level.


chrismayer (approved):
LGTM, thanks @geographika !

Issue #253, opened 2020-07-16
Drag and drop points with the Network Solver Tool


geographika:
This tool is a variant of the [DigitizeButtonController](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/button/DigitizeButtonController.js). 
Functionality could be created either as a subclass, a mixin to the current tool, or as configurable options on the current tool. 

It will allow a user to click points on the map to create a route as in the existing tool. Once the line has been created the tool will allow a user to select and drag the end points of the line, and then drop them. This will call the service again and return a new line feature. 

To add in conjunction with #88?

![image](https://user-images.githubusercontent.com/490840/87692323-eb487f80-c78b-11ea-91df-8e3ea52ebb6e.png)

simonseyock:
To achieve this and the delete button, I think it would be best to keep track of all drawn points and in which order they were drawn and which lines belong to which drawn point.

If this is done, it would actually also be possible to modify or delete points in between, not only the last point.

At the moment every time a new point is drawn all segments get recalculated. This actually is not necessary, or is it?

simonseyock:
Only the last part between the previous drawn point and the new point can change, is that true?

simonseyock:
After some thinking I came to the conclusion that recalculating everything should also be possible. This will result in more requests to the tool, though. What is the preferred way?

Is the network solver open source? Or is there a detailed documentation of the api?

simonseyock:
So I actually implemented it using the same method like before. You can ignore the above questions.

I did some refactoring of the class, though. Now the functions can be used in a more modular way which is useful for the modification.

Issue #252, opened 2020-07-16
Add Clear Features option to Network Solver


geographika:
Add a context menu option "Clear Features" to remove all lines and points created by the Network Solver tool.

![image](https://user-images.githubusercontent.com/490840/87691793-4b8af180-c78b-11ea-9f0b-00b5c5291742.png)

Use to clear active group only, or all features if 1 group.

simonseyock:
@geographika How can the network solver tool be accessed in the demo application?

geographika:
@simonseyock - this is only setup for roads in County Tipperary. Then use the Point tool to create paths. 

![image](https://user-images.githubusercontent.com/490840/96420760-30020d80-11f6-11eb-9ff7-f06c30950b90.png)


![image](https://user-images.githubusercontent.com/490840/96420699-195bb680-11f6-11eb-8271-1c87cad05bb6.png)

Issue #88, opened 2019-05-01
Remove Points with click on Network Solver Tool


geographika:
One piece of functionality that is missing is the ability to click on a point in the path to remove a node. This allows users to reroute with the tool without having to start from scratch again.
Example screencast from the old system: 

https://drive.google.com/file/d/12gVY-cUvKFZ9J568Vr1nQWsq7__qfqpw/view

I tried a couple of options:

- capturing the drawstart event and checking for features at the pixel
- a `ol.interaction.Modify` may be more appropriate

geographika:
Will need to check if this is still an issue in the current codebase

geographika:
Clicking on an existing point does not seem to clear the feature in the latest version of cpsi-mapview.

PR #317, opened 2020-10-19
Network solver tool updates

simonseyock:
This PR introduces the following features:
* A context menu entry to clear the active group
* The possibility to move any point in the point digitize mode
* The possibility to delete any point in the point digitize mode

It also contains a refactoring that was necessary to provide this features.

Resolves: #252
Resolves: #253
Resolves: #88

### DIFF
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -42,9 +42,19 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
 
     /**
      * OpenLayers modify interaction
-     * Used in polygon draw mode only
+     * Used in polygon and point draw mode
      */
     modifyInteraction: null,
+
+    /**
+     * OpenLayers pointer interaction for deleting points
+     */
+    deleteInteraction: null,
+
+    /**
+     * OpenLayers snap interaction for better vertex selection
+     */
+    snapInteraction: null,
 
     /**
      * CircleToolbar that will be set
@@ -78,6 +88,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
     onToggle: function (btn, pressed) {
         var me = this;
         var view = me.getView();
+        var modifiable = view.getModifiable();
 
         var deleteCondition = function (evt) {
             return ol.events.condition.platformModifierKeyOnly(evt) && ol.events.condition.singleClick(evt);
@@ -129,7 +140,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
         }
 
         // create the modify interaction
-        if (!me.modifyInteraction && type !== 'Circle') {
+        if (modifiable && !me.modifyInteraction && type !== 'Circle') {
             var modifyInteractionConfig = {
                 source: me.drawLayer.getSource(),
                 deleteCondition: deleteCondition
@@ -139,7 +150,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
             me.map.addInteraction(me.modifyInteraction);
         }
 
-        if (!me.deleteInteraction && type === 'Point') {
+        if (modifiable && !me.deleteInteraction && type === 'Point') {
             me.deleteInteraction = new ol.interaction.Pointer({
                 handleEvent: function (evt) {
                     if (deleteCondition(evt)) {
@@ -168,10 +179,10 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
 
         if (pressed) {
             me.drawInteraction.setActive(true);
-            if (type !== 'Circle') {
+            if (modifiable && type !== 'Circle') {
                 me.modifyInteraction.setActive(true);
             }
-            if (type === 'Point') {
+            if (modifiable && type === 'Point') {
                 me.deleteInteraction.setActive(true);
             }
             if (me.getView().getUseContextMenu()) {
@@ -183,10 +194,10 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
             }, 0);
         } else {
             me.drawInteraction.setActive(false);
-            if (type !== 'Circle') {
+            if (modifiable && type !== 'Circle') {
                 me.modifyInteraction.setActive(false);
             }
-            if (type === 'Point') {
+            if (modifiable && type === 'Point') {
                 me.deleteInteraction.setActive(false);
             }
             if (type === 'Circle' && me.circleToolbar != null) {
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -193,7 +193,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
             if (modifiable && type === 'Point') {
                 me.deleteInteraction.setActive(true);
             }
-            if (me.getView().getUseContextMenu()) {
+            if (me.getView().getGroups() || me.getView().getClearable()) {
                 me.map.getViewport().addEventListener('contextmenu', me.contextHandler);
             }
             // if another digitize button is pressed then this would come before the onToggle of the other button
@@ -212,7 +212,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
             if (type === 'Circle' && me.circleToolbar != null) {
                 me.removeCircleSelectToolbar();
             }
-            if (me.getView().getUseContextMenu()) {
+            if (me.getView().getGroups() || me.getView().getClearable()) {
                 me.map.getViewport().removeEventListener('contextmenu', me.contextHandler);
             }
 
@@ -253,6 +253,7 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
         evt.preventDefault();
 
         var me = this.scope;
+        var view = me.getView();
 
         var radioGroupItems = [];
         if (me.contextMenuGroupsCounter === 0) {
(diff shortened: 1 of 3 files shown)


simonseyock:
@geographika could you check if this provides the desired features? The delete is done by holding the ctrl key while clicking. It works the same with polygons.

The possibility to configure this is still missing. I would suggest one config option `featuresModifiable` that would turn on modification and deleting for both points and polygons.

Also circle are currently not possible to modify and it is not possible to remove a polygon entirely.

@geographika @chrismayer This PR is quite bulky to review. I suggest to either review the commit https://github.com/compassinformatics/cpsi-mapview/pull/317/commits/ef9b32455a2685b5da1aa4789177a3e7d78fdca4 beforehand or I can seperate this PR into 4 different units (clear group, refactoring, moving, deleting).

simonseyock:
To give a better visual indication of the delete it would be nice to have a special delete style in red that would be applied to the feature if hovered while the ctrl key is pressed. Is this desired?

simonseyock:
I think I made a conceptual error in understanding the network solver tool working with points. It calculates the shortest graph that contains all the given points. The order of the given points does not play a role. Is that correct?

geographika:
@simonseyock - the order is important - it solves from one point to the next based on the point `index`. Overlaps are then removed by the service. 
I've emailed @verendi with API doc links.

simonseyock:
@geographika thanks for the info. Based on that I was able to implement everything properly (At least I hope I did XD)

The functionality can already be reviewed, for the code I do some rebasing for easier reviewing

app/controller/button/DigitizeButtonController.js
```diff
@@ -243,21 +283,163 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
...
+     */
+    handlePointDelete: function (evt) {
+        var me = this;
+
+        var features = me.map.getFeaturesAtPixel(evt.pixel, {
+            layerFilter: function (layer) {
+                return layer === me.drawLayer;
+            }
+        });
+        if (features.length) {
```
geographika, line 419: `features` could be null here and throws an error (happened during testing)

app/controller/button/DigitizeButtonController.js
```diff
@@ -243,21 +283,163 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
...
+    handleModifyEnd: function (evt) {
+        var me = this;
+        var view = me.getView();
+
+        // find modified feature
+        var drawFeature = me.map.getFeaturesAtPixel(evt.mapBrowserEvent.pixel, {
+            layerFilter: function (layer) {
+                return layer === me.drawLayer;
+            }
+        })[0];
```
geographika, line 366: `drawFeature` could be null (happened during testing)

geographika (commented):
@simonseyock - thanks for this - it provides nice improvement to the UX. 

Some comments:

- the context-menu should always be available with a "Clear All" option. When groups are used the "Clear All" option can remain at the bottom to remove all features from all groups. "Clear All" should be available for all the different types (polygon, circle, line). 

image

The circle tool keeps adding new features. Maybe the code at #313 needs to be (re)applied?
image

I'm not sure I fully understand the deletion process. In the screencast at https://drive.google.com/file/d/1oPuIprkI4k9j_uDd-qqtd0gG7IldOObk/view I held CTRL and clicked the point but the whole path was removed. I was using Chrome on Windows if that makes a difference. 
In terms of approach, would it be possible to add a "Delete Point" option to the context menu when right-clicking on a point? Then the deletion could be controlled with just the mouse.

app/controller/button/DigitizeButtonController.js
simonseyock: did this happen during deletion of a polygon vertex or any other modification? If it happened during any other modification I have to investigate further.

simonseyock:
I fixed the deletion and the other points.
Only for the context menu I need a clarification: At the moment it is controlled by the `useContextMenu` option. Should this option get removed then?

simonseyock:
I just keep the option for now.

simonseyock:
Should the groups also be available for circles and polygons? If that is the case I would suggest two options `groups` and `clear` instead of the option `useContextMenu`.

Also there is the option `clearDrawnFeature` which does not make really sense in my opinion. In which case would you want to keep your drawn features? I would remove it because it is not really clear what it would do for example in case of modified or deleted points.

geographika:
> 
> 
> I fixed the deletion and the other points.
> Only for the context menu I need a clarification: At the moment it is controlled by the `useContextMenu` option. Should this option get removed then?

Yes this must seem a little confusing. Only the point solver uses groups, which is set by `useContextMenu`. It would make more sense to have a separate `useGroups` option.

geographika:
> Should the groups also be available for circles and polygons? If that is the case I would suggest two options `groups` and `clear` instead of the option `useContextMenu`.

`groups` and `clear` sound a good approach.
Groups are not used for circles and polygons currently. 

> Also there is the option `clearDrawnFeature` which does not really make sense in my opinion. In which case would you want to keep your drawn features? I would remove it because it is not really clear what it would do for example in case of modified or deleted points.

In `CpsiMapview.view.toolbar.MapTools` you'll see this option is set to `false` for polygons and circles so the feature is retained after the lines are returned. 

image

simonseyock:
> In CpsiMapview.view.toolbar.MapTools you'll see this option is set to false for polygons and circles so the feature is retained after the lines are returned.

Ok I thought this was the default. What should the option do for points then? Behave like modifiable was set to false?

geographika:
> 
> 
> > In CpsiMapview.view.toolbar.MapTools you'll see this option is set to false for polygons and circles so the feature is retained after the lines are returned.
> 
> Ok I thought this was the default. What should the option do for points then? Behave like modifiable was set to false?

Not sure I fully understand the question. 
It is just the point tool that needs to hide the "drawn" points as they are also returned by the service.

geographika:
@simonseyock - let me know if the above answer your questions. I tried the latest branch but this doesn't seem to have addressed the comments in https://github.com/compassinformatics/cpsi-mapview/pull/317#pullrequestreview-513569640 - I'm not sure if these have been added or are yet to be pushed.

simonseyock:
@geographika the context menu is not modified yet, because I had these questions. The circles should work, though.

About the options: I am in general way more in favour of providing only options that are really needed because with every new option you generate new edge cases where different options are combined that you did not think of beforehand. That's why asked all these questions :) In general I think sensible defaults are preferable. But for now I will implement with the 3 options.

geographika:
There are 3 variations of the tool. Each tool involves 2 layers:

1. The user "draw" layer, which contains the features created by the user
2. The "results" layer which contains line features sent back from the server. 

The 3 tools are:

- the point solver tool. Any of the points created by the user in the "draw" should not be retained once the "results" layer is populated. Any previous "results" should be cleared, unless they are in another group. This is the only solver which works with groups at the moment. 
- the polygon solver tool. With this tool the last polygon created by the user in the "draw" layer should be retained to show the outline around the "results". Every time a user creates a new polygon both the "draw" layer and "results" layer should be cleared. 
- the line solver currently has no implementation and is not used. 

If only 2 options are required to implement these then all the better.

simonseyock:
@geographika  You are forgetting the circle selection.

In my opinion there is no need for any of theses options - everything could depend on the tool type:
If it is `Point` provide groups, a `clear active group` option and make them modifiable - because why not? The modify interaction should not interfere with the normal usage.
If it is `Circle` or `Polygon` provide a `clear all` option. If it is `Polygon` also make it modifiable (this is currently the behaviour).

simonseyock:
And always do not clear the drawn feature for `Circle` or `Polygon`

simonseyock:
I added a commit that provides the options, though.

geographika:
> 
> 
> @geographika You are forgetting the circle selection.
> 
> In my opinion there is no need for any of theses options - everything could depend on the tool type:
> If it is `Point` provide groups, a `clear active group` option and make them modifiable - because why not? The modify interaction should not interfere with the normal usage.
> If it is `Circle` or `Polygon` provide a `clear all` option. If it is `Polygon` also make it modifiable (this is currently the behaviour).

Yes the options aren't really required - all is driven by the tool type. Sorry I missed the Circle tool - this should behave like the polygon tool. All but the last entered user feature are removed. 
Yes all features should be modifiable.

simonseyock:
@geographika Just to be clear: Should I then remove the options `useContextMenu`, `clearDrawnFeature` and also not add the new options `modifiable`, `groups`, `clear`?

geographika:
> 
> 
> @geographika Just to be clear: Should I then remove the options `useContextMenu`, `clearDrawnFeature` and also not add the new options `modifiable`, `groups`, `clear`?

I think defaults can be used based on the tool type rather than having options - with the exception of `goups` - this would apply to the point solver only.

simonseyock:
@geographika The options are removed, all your remarks should be resolved.

@chrismayer Could you review this code? It maybe makes sense to review the commit https://github.com/compassinformatics/cpsi-mapview/pull/317/commits/d7a65654c80613f125d5e324df3603f126de785c first which should not add any new functionality and the rest afterwards: https://github.com/compassinformatics/cpsi-mapview/pull/317/files/d7a65654c80613f125d5e324df3603f126de785c..bee8297cdbbc42ec6182eaac3a083a25b7815afb

chrismayer (approved):

geographika:
@simonseyock - works very well. Thanks for all the updates. I'm happy to merge this (I see @chrismayer has approved) so will do so in the next hour.

Issue #304, opened 2020-10-06
Allow WMS Style Selection from a context menu


geographika:
Currently, styles are added as radio buttons to a LayerTreeNode:

![image](https://user-images.githubusercontent.com/490840/95223236-a451af80-07f9-11eb-8125-0f35c2d5e7db.png)

It would be useful for some applications to allow this switching via a context menu, with an expandable menu time listing all the styles, and allowing selection. 
It would also help mitigate the performance issues noted in #274 and jumping in #287. 

![image](https://user-images.githubusercontent.com/490840/95223335-bf242400-07f9-11eb-8890-ca3698b931ed.png)

The selected style should be added to the layer's text in the menu item in brackets e.g. `GAS WFS (CUSTOM TITLE)`. 

Ideally the current radio button approach would continue to work as present, but a config option would allow setting as a context menu. 

`CpsiMapview.plugin.TreeColumnStyleSwitcher` would be refactored and could be added to a `CpsiMapview.plugin.TreeColumnContextMenu`.

PR #319, opened 2020-10-20
Enable switchlayer change in subfolder

### LINKED ISSUE #265: switchlayer not changing when in a subfolder
The following layer search code will need to be made to be recursive to check for overlays within overlays: https://github.com/compassinformatics/cpsi-mapview/blob/master/app/factory/Layer.js#L926

JakobMiksch:
Resolves #265 

Warning: the diff from GitHub looks horrible. 
Actually, only the for-loop has been converted to a recursive function. Hence, the real change is not that big.

I noticed the layertree does not properly update the switch layer anymore. This seemed to be introduced with https://github.com/compassinformatics/cpsi-mapview/commit/081b0e71ca1d1e729bfedaf2cc964f6723291f1e

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -914,87 +914,110 @@ Ext.define('CpsiMapview.factory.Layer', {
      */
     handleSwitchLayerOnResolutionChange: function(evt) {
         var resolution = evt.target.getResolution();
-        var allLayers = BasiGX.util.Map.getMapComponent().getMap().getLayers();
-        var overlayGroup = BasiGX.util.Layer.getLayerByName('Layers', allLayers);
+        var layerCollection = BasiGX.util.Map.getMapComponent().getMap().getLayers();
 
-        if (!Ext.isDefined(overlayGroup)) {
-            return;
-        }
-        var overlayCollection = overlayGroup.getLayers();
+        LayerFactory.checkSwitchLayersRecursively(layerCollection, resolution);
+    },
 
-        overlayCollection.forEach(function (layer, index) {
+    /**
+     * Traverses layertree including subgroups and
+     * changes switch layers if necessary
+     * @param {*} overlayCollection the layers of the map
+     * @param {*} resolution the current resolution of the map
+     */
+    checkSwitchLayersRecursively: function(overlayCollection, resolution) {
 
-            if(layer.get('isSwitchLayer') && LayerFactory.isLayerSwitchNecessary(layer, resolution)){
+        overlayCollection.forEach(function (layerOrGroup, index){
 
-                var switchConfiguration = layer.get('switchConfiguration');
-                // restore current layer visibility
-                switchConfiguration.visibility = layer.getVisible();
-                // also apply current filter and selected style
+            if(layerOrGroup instanceof ol.layer.Layer){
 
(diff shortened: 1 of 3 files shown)


chrismayer:
> I noticed the layertree does not properly update the switch layer anymore. This seemed to be introduced with 081b0e7

Can confirm this. Seems like an unrecognized side effect of #289. I'll have a deeper look at this but surely unrelated to this PR.

chrismayer:
@JakobMiksch The tree node should update properly with the fix provided in #322 (please test). Could you add this fix to your PR? Or how should we handle that?

JakobMiksch:
@chrismayer yes, I am trying to include your changes into this pull request as well. Has not worked yet, but I am tracing the problem ...

chrismayer:
@JakobMiksch I created a PR against your branch for this: https://github.com/JakobMiksch/cpsi-mapview/pull/1 Please review and check if this does the trick for you.

JakobMiksch:
@chrismayer unfortunately it does not work yet. The fix works on the top level, but in subgroups it only changes the label of the layertree once, and then not anymore at all.

JakobMiksch:
I merged the latest changes of the `master` - now the functionality works like expected. I tested it for the case that the switch layer is on top level, in a group and even in a sub-group of a group.

app/factory/Layer.js
```diff
@@ -914,87 +914,110 @@ Ext.define('CpsiMapview.factory.Layer', {
...
-                    var wmsFilterString = wmsFilterUtil.getWmsFilterString(newLayer);
+                    if (newLayer.get('isWms')) {
 
-                    // apply new style parameter and reload layer
-                    var newParams = {
-                        STYLES: activeStyle,
-                        FILTER: wmsFilterString,
-                        cacheBuster: Math.random()
-                    };
+                        activeStyle = LegendUtil.getWmsStyleFromSldFile(activeStyle);
```
geographika, line 967: Error thrown here as `activeStyle` is undefined. This is when the application starts zoomed in, and the switch layer is off by default, and then zoomed out so the WMS takes over. This happens if all switch layers are switched off and the map is simply zoomed in and out.

geographika:
Thanks @JakobMiksch 

I've tested locally and integrated into a larger project. 

One minor issue in the `mapview` testing - the style radio buttons is not always set/retained when switching layers:

image

Maybe less of an issue if #304 is implemented. 

In the larger project I get an error when zooming in and out. It could be related to the same issue above - the WMS style is not set when the application is started up and the map zoomed in.

chrismayer:
Thanks for your feedback @geographika. I'd pledge to merge this in since this works as expected and create a separate ticket for the remaining minor issue described by you above. Otherwise things feeling a bit messed up because this brings a larger change (larger than expected initially) and maybe has other side-effects with #304.We surely will have a look at the new issue. What do you think?

geographika:
@chrismayer - yes happy to merge. I've never switched branches as much as this week!

chrismayer:
> @chrismayer - yes happy to merge. I've never switched branches as much as this week!

:rofl: : same for me :pray: 

I'll merge then...thanks again for @JakobMiksch and @geographika!

PR #322, opened 2020-10-21
Ensure tree node text is updated for switch layer

chrismayer:
This ensures that for an exchanged switch layer the correct text (layer name) will be shown in the tree node.

Follow-up for #289 as detected in #319.


geographika (approved):
Seems good as far as I can tell!

JakobMiksch (commented):
Tested it and it does what it should. I tried to combine it with #319 but not successful yet.

chrismayer:
Okay, then I'll merge this. The rest should be done in #319.

Issue #245, opened 2020-07-13
Feature Selection Tool


geographika:
Allow features to be selected in a layer by clicking with a tool linked to a WFS grid. The tool should be available on the [grid window tools](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/grid/Grid.js#L90):

![image](https://user-images.githubusercontent.com/490840/87304661-72d88900-c515-11ea-85a3-c99a89e0d26c.png)

A user would be able to click on features in the map and they would be visible in the grid. SHIFT could be used to select multiple features. 

This selection can be independent of #244, as the WFS grid has its own vector layer (see [vectorLayerKey](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/model/grid/Grid.js#L18) and [vectorLayer](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/grid/Grid.js#L147). 

The selection tool does need to consider how to interact with other filters on the grid. A grid can have:

- a [spatial filter](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/grid/Grid.js#L206)
- [attribute filters](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/grid/Grid.js#L124) from the grid

The selection tool could collect feature Ids from the visible map layer and pass these as an additional filter to the WFS. Alternatively the Ids could override any other selection. 

The WMS should be kept in synch with the grid.

chrismayer:
General approach is clear, but I see the following edge case: 

- The user selects several features by polygon (let's say 50 features). 
- The grid shows only the first 20 features due to paging. 
- The user clicks a feature on the map which is not on grid page 1

How should we handle that?

geographika:
@chrismayer  - a few initial comments - I'm going to check again how this is handled in various GISs. 
The tool has 2 main purposes, so maybe the functionality should be set with a dropdown allowing the user to switch? 
This is similar to how QGIS and [ArcMap ](https://pro.arcgis.com/en/pro-app/help/mapping/navigation/selection-options.htm) work. 

1. **New Selection** - in this case clicking on a feature should override any other filters and add a single feature to the grid. So after one selection one feature would be present in the grid. Not sure if holding SHIFT and selecting another feature would add a second feature to the grid. 
2. **Add to Selection** - in this case any features not already in the grid would be added to the filters. Depending on the sort order it may or may not be visible in the grid. So in the example above there would be 51 total records to page through. 



[...]

PR #323, opened 2020-10-21
Add FeatureSelection tool

chrismayer:
This adds a FeatureSelection tool as described in #245.

245-feature-select


chrismayer:
Some remarks: 

- Due to clustering of the vector features it sometimes looks like just one feature was clicked but in the grid two or more appear. 
- Would it be useful to inform the user if no feature was hit by the map click? Currently anything is done. Maybe a little feedback would increase usability?

simonseyock:
As long as your interaction is active you can disable the default behaviour by calling
```
olMap.set('enableDefaultClick', false);
```
and enable it again with setting it to true.

app/controller/button/FeatureSelectionButton.js
```diff
@@ -0,0 +1,206 @@
...
+        var me = this;
+
+        if (!me.modeSelector) {
+            me.modeSelector = Ext.create('Ext.button.Split', {
+                text: 'Add to selection',
+                hidden: true,
+                menu: new Ext.menu.Menu({
+                    items: [
+                        {
+                            text: 'Add to selection',
```
geographika, line 100: Change to "Add to Selection"

app/controller/button/FeatureSelectionButton.js
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * Adds the mode selector UI to the toolbar (if not existing) and shows it.
+     */
+    addModeSelectorUi: function () {
+        var me = this;
+
+        if (!me.modeSelector) {
+            me.modeSelector = Ext.create('Ext.button.Split', {
+                text: 'Add to selection',
```
geographika, line 95: Change to "Add to Selection"

app/controller/button/FeatureSelectionButton.js
```diff
@@ -0,0 +1,206 @@
...
+                    items: [
+                        {
+                            text: 'Add to selection',
+                            handler: function (menu) {
+                                me.filterMode = 'ADD_TO_SELECTION';
+                                me.modeSelector.setText(menu.text);
+                            }
+                        },
+                        {
+                            text: 'New selection',
```
geographika, line 107: Change to "New Selection"

geographika (commented):
@chrismayer - playing with this now! Looks excellent. 
I'll need some time to try and test integration in the larger application.

A couple of initial comments. 

I might change the `selModel` on the grid to `MULTI` as then you can easily highlight all features in the grid on the map. Any foreseeable problems with this?

I think the clustering issue is fine.

In the old system when a feature was not selected I had an Ext animation flash a message in a status bar at the bottom of the screen. Maybe if an event was fired this functionality could be added in a future release.

app/controller/button/FeatureSelectionButton.js
chrismayer: Done in 4f03a22

app/controller/button/FeatureSelectionButton.js
chrismayer: Done in 4f03a22

app/controller/button/FeatureSelectionButton.js
chrismayer: Done in 4f03a22

app/controller/button/FeatureSelectionButton.js
```diff
@@ -0,0 +1,206 @@
...
+
+        // clear FIDs to filter when we have a new selection
+        if (me.filterMode === 'NEW_SELECTION') {
+            me.fidsToFilter = [];
+        }
+
+        // collect all IDs of clicked features
+        me.map.forEachFeatureAtPixel(evt.pixel, function(feature, layer) {
+            // add check for correct layer
+            if (layer.id === view.queryLayer.id) {
```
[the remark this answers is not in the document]
chrismayer: Done in 4f03a22

chrismayer:
I addressed the review comments of @geographika.

Also removed the conflict with the GetFeatureInfo tool and added some behaviour if no feature was hit by clicking on the  map:

  - The view fires an event `'cmv-no-feature-selected'` 
  - By setting the views config option `showNoSelectionMessage` to `true` a little `Toast` is shown in the lower-right corner of the app with a message

chrismayer:
> I might change the selModel on the grid to MULTI as then you can easily highlight all features in the grid on the map. Any foreseeable problems with this?

I don't think so because this does not interact with the selection model at all.

geographika:
@chrismayer - I may make a follow-up pull request to allow a grid to have just the "Select by Polygon" or both select tools - but please merge this as-is - thanks.

PR #324, opened 2020-10-21
Add metadata plugin

### LINKED ISSUE #240: Layer metadata tree plugin
Create a new [tree plugin](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/plugin/TreeColumnContextMenu.js) that adds a new "Layer Metadata" menu option. 

When clicked this will do the following:

1. Create request based on layer name (assume all layers have metadata rather than requiring a flag in layers.json)
2. Get metadata XML from server
3. Populate grid based on metadata name/value pairs and open in window

MapServer is used as the backend. 
GetMetadata support was implemented in https://mapserver.org/development/rfc/ms-rfc-82.html
A sample request is: http://my.host.com/cgi-bin/mapserv?request=GetMetadata&layer=foo

MapServer’s layer API supports the [ISO 19115:2003 geospatial standard](https://www.iso.org/standard/26020.html)

JakobMiksch:
resolves #240 

It shows the "Metadata" option for all layers, like specified in the description. 

However, metadata can only be retrieved for WMS and WFS layers from MapServer. 

Metadata for Vectortiles could also be possible, but the layername and the baseurl are difficult to extract, since they are not separetely added in the config file (like for WMS and WFS). Both layername and baseurl could of course be extracted from the connection-url:
`https://w08-mapserver.compass.ie/mapserver/?map=/MapServer/apps/mapview-demo/example.map&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=waterways&TILED=false&CRS=EPSG%3A3857&WIDTH={width}&HEIGHT={height}&BBOX={bbox}&STYLES=Waterways&FORMAT=mvt`
However, it might be more elegant and consistent to provide the layername and the baseurl directly in the config file. 

If no metadata is available, an alert window is shown. This is probably not the behavior you would like for long term.
@geographika what is your preferred way to display "no metadata"?

### DIFF
--- a/app/view/menuitem/LayerMetadata.js
+++ b/app/view/menuitem/LayerMetadata.js
@@ -1,0 +1,173 @@
+/**
+ * MenuItem to show a metadata window for a layer
+ *
+ * @class CpsiMapview.view.menuitem.LayerMetadata
+ */
+Ext.define('CpsiMapview.view.menuitem.LayerMetadata', {
+    extend: 'Ext.menu.Item',
+    xtype: 'cmv_menuitem_layermetadata',
+    requires: [
+        'CpsiMapview.util.Layer'
+    ],
+
+    /**
+     * The connected layer for this item.
+     *
+     * @cfg {ol.layer.Base}
+     */
+    layer: null,
+
+    /**
+     * Text shown in this MenuItem
+     * @cfg {String}
+     */
+    text: 'Metadata',
+
+    /**
+     * @private
+     */
+    initComponent: function () {
+        var me = this;
+
+        me.handler = me.handlerFunc;
+
+        me.callParent();
+    },
+
+    /**
+     * Executed when this menu item is clicked.
+     * Opens a window with metadata for the connected layer.
+     */
+    handlerFunc: function () {
+        var me = this;
+
+        /**
+         * Helper function to notify user
+         * that metadata is not availble
+         */
+        function alert_no_metadata(){
+            alert('No metadata available.');
+        }
+
+        // retrive layername for different types of layers
+        var layerName;
+        if(me.layer.get('isWfs')){
+            layerName = me.layer.get('featureType');
+        }
+        else if(me.layer.get('isWms')){
+            layerName = me.layer.get('layers');
+        }
+        else{
+            // Currently only metadata from WMS and WFS can be received
+            // because other layers do not come from MapServer
+            // VectorTiles are an exception: the config does
+            // not provide easy access to name and baseurl
+            alert_no_metadata();
+
+            return;
+        }
+
+        // build url to Metadata XML
+        var baseurl = me.layer.get('url');
+        var requestUrl = baseurl + '&REQUEST=GetMetadata' + '&layer=' + layerName;
+
+        Ext.Ajax.request({
+            url: requestUrl,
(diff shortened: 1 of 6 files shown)


app/view/menuitem/LayerMetadata.js
```diff
@@ -0,0 +1,173 @@
...
+     * Opens a window with metadata for the connected layer.
+     */
+    handlerFunc: function () {
+        var me = this;
+
+        /**
+         * Helper function to notify user
+         * that metadata is not availble
+         */
+        function alert_no_metadata(){
```
chrismayer, line 48: Please move this outside of handlerFunc as member function. Also please use camel case notation

chrismayer (approved):
LGTM thanks, I added a minor thing. Once @geographika gave his feedback this ist good to go IMHO after you made Travis happy ;)

app/view/menuitem/LayerMetadata.js
JakobMiksch: fixed with 8e7108b5c8286adfce0795e7c372108b3748adf2

JakobMiksch:
this commit 5755e42eeb6ae3787d7aee17f8e2028ef79f098d makes ESLint and therefore Travis and Azure happy

app/view/menuitem/LayerMetadata.js
```diff
@@ -0,0 +1,171 @@
...
+                        source['keywords'] = keywordString;
+                    }
+
+                    // case no metadata could be extracted
+                    if(Ext.Object.isEmpty(source)){
+                        me.alertNoMetadata();
+                    }
+                    // metadata is fine and can be displayed
+                    else{
+                        var windowTitle = 'Metadata '+ me.layer.get('name');
```
geographika, line 129: Would read better as '{layername} Metadata'

geographika:
@JakobMiksch - works great - thanks!

A few comments:

- I presume the ISO schemas can be downloaded and hosted for the project if required without any issues?
- To handle missing metadata I'd suggest the Metadata context-menuitem is only shown for a layer with a `hasMetadata: true` flag in the config. This would be easy to set for all layers of a certain type such as WMS/WFS using the type defaults here: https://github.com/compassinformatics/cpsi-mapview/blob/2410393dadaadc783eb51e3e1fb7006a093e0297/resources/data/layers/default.json#L22
- If MD is still missing an ExtJS message box with a warning should be displayed
- For vector tiles setting `layername` and the `baseurl` directly in the config file sounds a good approach. It would just require setting these properties in the `default.json`?

JakobMiksch:
@geographika I addressed your remarks:

- yes, the ISO schema can also be downloaded - maybe @chrismayer can elaborate further if necessary
- I added the flag `hasMetadata: true`
- the alert for missing metadata is replaced with a Ext message box
- the vector tile config now has the flags `baseUrl` and `layerIdentificationName`. The rather long name `layerIdentificationName` was intentionally chosen in order to avoid confusions with flag `name` which is actually the title
- I added `CpsiMapview.view.menuitem.LayerMetadata` to the "requires" of the layer tree component

geographika:
Thanks @JakobMiksch !

Issue #287, opened 2020-09-30
Layer Tree scroll bar jumps when interacting with layers


select-8:
When you scroll down to one of the layers lower in the tree and turn it on or off, the layer tree jumps back up (see attached gif).
![creen_Recording_30-09-2020_16-14-37_](https://user-images.githubusercontent.com/47751242/94705766-52acaf00-0339-11eb-95a4-ba84b386b5b3.gif)

geographika:
Possible solution is the [preserveScrollOnRefresh](https://docs.sencha.com/extjs/6.7.0/classic/Ext.view.AbstractView.html#cfg-preserveScrollOnRefresh) option.

geographika:
@chrismayer - sort-of related to the issues in #274 - putting a debugger in `getLegendHtml` in `CpsiMapview.plugin.BasicTreeColumnLegends` shows the tree completely disappears and is redrawn from scratch. 

Setting the following [here](https://github.com/compassinformatics/cpsi-mapview/blob/77c824fdb93969ee1333cb14971f5b7d5c4053d8/app/view/LayerTree.js#L20) doesn't appear to have an effect - the scroll position is lost:

```js
    preserveScrollOnRefresh: true,
    preserveScrollOnReload: true,
```

geographika:
Additional screencast at https://drive.google.com/file/d/1q0VLeVi9C8KPWWN2bJkRrM1VbzbpbYgQ/view
When "A WMS 4" is switched on the layer tree jumps back to the top.

PR #325, opened 2020-10-22
Make style switcher usable in LayerTree context menu

### LINKED ISSUE #304: Allow WMS Style Selection from a context menu

chrismayer:
This makes the RadioGroup to switch between styles (normally rendered under the corresponding layer tree node) usable in the context menu of the tree node. By setting the config property `styleSwitcherBelowNode` to false for the LayerTree will force the rendering in the context menu. Setting it to `true` (default) will enforce the old behavior.

This is related to #274 since this should bring better perfomance for the LayerTree since the radios are not rendered directly at the node and the update frequency should be lower. Also should improve the jumping behavior as reported in #287.   

This also ensures that the activated style is shown next to the layer name in the layer tree node for the case that the style switcher is only available in the context menu and not permanently visible.

304-style-radio-ctx-menu

Closes #304.

### DIFF
--- a/app/data/model/LayerTreeNode.js
+++ b/app/data/model/LayerTreeNode.js
@@ -26,6 +26,41 @@ Ext.define('CpsiMapview.data.model.LayerTreeNode', {
             name: '__toggleMode',
             type: 'string',
             defaultValue: 'ol3'
+        },
+        {
+            // overwrite the text property to add custom behaviour
+            name: 'text',
+            type: 'string',
+            persist: false,
+            convert: function(v, record) {
+                if (!v) {
+                    // folders / LayerGroups
+                    var layerGroup = record.getOlLayer();
+                    if (layerGroup) {
+                        return record.getOlLayer().get('name');
+                    }
+                }
+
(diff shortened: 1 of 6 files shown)


geographika:
Thanks @chrismayer - works perfectly, and tested on a couple of applications. 
No performance issues now with the layer tree. 
Jumping only happens when collapsing nodes not expanding - I'll investigate further in the future, but not an issue for now. 

Please merge!

PR #327, opened 2020-10-22
Miscellaneous fixes

geographika:
Various fixes for issues discovered when integrating all tools into a larger application. 
Mainly relating to requires and making sure `defaultClickEnabled` is set when using various tools. 

@JakobMiksch  - I added 41e74a2 as a temp fix for errors thrown in the layer switcher. 

Merging all now to give a demo of all the latest features but would appreciate a quick review.


chrismayer:
Even tough it is merged already: This looks good to me. No remarks. Thanks @geographika !

PR #328, opened 2020-10-22
Minor FeatureSelectionButtonController fixes

geographika:
Avoid errors if the queryLayer is not set. 
Rename file to match controller name so it is loaded correctly. 
Increase setTimeout to `100` so when used in a `segmentedbutton` there is no race condition that reactivates map tools.


chrismayer (approved):
LGTM, thanks @geographika!

PR #330, opened 2020-10-23
Allow use of a default map if supplied in a derived class

geographika:
This allows reuse of all the code previously in `items` for any derived applications.


chrismayer:
Just to clarify: For the case you derive and you have to overwrite the `items` property? Otherwise the usage of the default map should also be possible in derived classes.

PR #334, opened 2020-11-02
Expand panel plugin

### LINKED ISSUE #321: Create a Panel Expander Mixin
Create a reusable mixin or plugin that allows a panel to be expanded to fill the viewport, or a parent component, such as the centre region of the viewport.

This would work in the same way as the "Maximize example code content" button found in the ExtJS kitchen sink example: https://examples.sencha.com/extjs/6.7.0/examples/kitchensink/?classic#window-variations

image

This does not appear to be a part of a framework, but instead uses a "ghost" panel and fills this with the component when the "Maximize" button is clicked. 

A Sencha Fiddle link with similar functionality: https://fiddle.sencha.com/#view/editor&fiddle/1juu

simonseyock:
Resolves #321

### DIFF
--- a/app/plugin/ExpandPanel.js
+++ b/app/plugin/ExpandPanel.js
@@ -1,0 +1,89 @@
+Ext.define('CpsiMapview.plugin.ExpandPanel', {
+    extend: 'Ext.plugin.Abstract',
+    alias: 'plugin.cmv_expand_panel',
+
+    layout: 'fit',
+
+    /**
+     * This is the parent panel that the expanded panel will be reinserted after maximizing. By default is set to the
+     * direct parent. A parent panel needs to exist for this plugin to work.
+     * If a window should be made expandable use the config `maximizable` instead.
+     */
+    parentPanel: null,
+
+    /**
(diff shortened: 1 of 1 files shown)


geographika:
@simonseyock - thanks for this. I've tried it in the application and it successfully adds to the component to the maximized window. 
There are a couple of issues such as there is no controller, so buttons etc. cause errors, and the grid doesn't expand to fill the available space. 

Rather than trying to solve this generically, which would likely be very complicated and still have issues, could we add two events - `expanded`, `collapsed` to the plugin, that would fire an event with the `component` and `parent` (`parentPanel` and `cmp` in the code)? Then I can handle things like resizing, hiding titles, setting controllers etc. on a case-by-case basis.

simonseyock:
@geographika I created `expand` and `restore` events. The events will fire on the expanded/restored panel and on it's parent and will pass the targeted child component as the first argument to the event listener. This way you can also distinguish if the component itself expanded or if it was a child (compare `this` or `me` to the passed component).

geographika:
Excellent - works perfectly and I can handle hiding/showing the title bar with those events in the application. 
Thanks @simonseyock

PR #335, opened 2020-11-11
Expander ExtJS bug workaround

geographika:
Add empty config as workaround for Cannot read property viewModel of null. 

See https://forum.sencha.com/forum/showthread.php?469880-Cannot-read-property-viewModel-of-null
Creating an empty config object on the window prevents this error.


geographika:
Error still occurs so using override as mentioned in related link to forum.

PR #337, opened 2020-11-12
please ignore me

leonardospina:
Some users experienced an issue where the password manager of their browser would fill in the credentials for them, but the UI wouldn't react properly since no "event" (typing) actually occurred on the form, hence the native validation wouldn't kick in and the "Login" button wouldn't be enabled for them.


leonardospina:
wrong check in

PR #351, opened 2020-12-09
Only set defaultClickEnabled to true if a tool has been activated

geographika:
Currently if a button is on a form, and is never activated, if the form is closed then the `defaultClickEnabled` is set back to true, even if another tool is active. 
The feature info tool then becomes active along with another active tool.
This pull request ensures the `defaultClickEnabled` is only set when the tool was actually active to avoid this. 

**TODO** there is still an issue that buttons are in a toggle group so activating one sets `defaultClickEnabled` to true which then triggers the `onToggle` of another button which sets it back to false.

### DIFF
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -198,7 +198,15 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
                 me.map.set('defaultClickEnabled', false);
             }, 0);
         } else {
+            // only reset the defaultClickEnabled on the map if the tool
+            // was already active - as this function is called in onBeforeDestroy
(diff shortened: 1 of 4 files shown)


geographika:
@simonseyock - any ideas on how to handle the `onToggle` issue greatly appreciated!
Maybe setting a higher timeout would work?

simonseyock:
@geographika can you tell me how I can reproduce the problem in the default test application?

simonseyock:
I changed the implementation so it counts the enabled / disabled calls. this should be better than using timeouts. I did not test if it does work correctly. 

See https://github.com/simonseyock/cpsi-mapview/tree/defaultClickEnabled-fix

geographika:
Superseded by #362

PR #353, opened 2020-12-10
Only show Gazetteer dropdown when text is entered

geographika:
Avoid an empty loading store when clicking the dropdown arrow.

image

### DIFF
--- a/app/view/combo/Gazetteer.js
+++ b/app/view/combo/Gazetteer.js
@@ -46,14 +46,22 @@ Ext.define('CpsiMapview.view.combo.Gazetteer', {
             })
         });
 
-        me.on('beforequery', function () {
-            var val = this.getRawValue();
-            var url = me.url + encodeURIComponent(val.trim());
-            me.getStore().getProxy().setUrl(url);
-        });
-
-        me.on('afterrender', function () {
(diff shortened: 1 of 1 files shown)


app/view/combo/Gazetteer.js
```diff
@@ -46,17 +46,35 @@ Ext.define('CpsiMapview.view.combo.Gazetteer', {
...
-            me.locationLayer.set('displayInLayerSwitcher', false);
+        me.on({
+            beforequery: function () {
+                var val = this.getRawValue();
+                // set a key on the layer using the unique id of the combo
+                me.locationLayer.set('layerKey', me.id);
+                if (Ext.isEmpty(BasiGX.util.Layer.getLayersBy('layerKey', me.id))) {
+                    // if the layer is not yet in the map then add it
+                    if (me.map) {
+                        me.map.addLayer(me.locationLayer);
```
chrismayer, line 59: Are there any conflicts with the adding the layer adding to the map by the base class?

chrismayer: Oh, OK. Then it is good to go. Thanks for clarification.

chrismayer (approved):

PR #356, opened 2020-12-21
Points are now clickable

### LINKED ISSUE #348: Allow point solver to add points on existing feature
Currently, clicking on an existing point won't trigger a solve request. This means a full circle cannot be created (which is useful for drawing features on roundabouts). 

image

See video at https://drive.google.com/file/d/1Rl5sfH_lFTIk18mLVs4rn1kzzuY3EslP/view

This related to the [DigitizeButtonController](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/button/DigitizeButtonController.js). 

Unclear if a handler has to be added to the Draw or Modify interactions.

simonseyock:
Resolves #348

### DIFF
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -365,6 +365,8 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
                 return;
         }
 
+        evt.stopPropagation();
+
         resultPromise
             .then(me.handleFinalResult.bind(me))
             .then(undefined, function (err) {
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -81,6 +81,11 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
     contextMenuGroupsCounter: 0,
 
     /**
+     * Determines if event handling is blocked.
+     */
+    blockedEventHandling: false,
+
+    /**
      * Main handler which activates or deactives the interactions and listeners
      * @param {Ext.button.Button} btn The button that has been pressed
      * @param {boolean} pressed The toggle state of the button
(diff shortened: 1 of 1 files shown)


geographika:
@simonseyock - thanks for this. There seems to be an issue where each click results in two requests: the  `handleDrawEnd` calls `doAjaxRequest`, and then `handlePointClick` is fired - the new point has been added to the layer and so a second `doAjaxRequest` is triggered. I'll try and take a screencast later.

simonseyock:
@geographika the added line should solve the issue

geographika:
@simonseyock - unfortunately the `handlePointClick` is still called even with `evt.stopPropagation()`. Were you able to recreate?

simonseyock:
I somehow thought I saw only one request but you are right, there are two.

simonseyock:
I am sorry. I look into it again.

simonseyock:
I solved this issue with using a timeout of 300 ms. I do not like it very much but openlayers does not offer a good way to do this here.

geographika:
@simonseyock - thanks. This fixes the double request, but unfortunately breaks the context-menu - right-clicking now also triggers a request rather than showing the menu.

simonseyock:
Yes I will fix that. The handling is a little bit of a mess. One thing is that the draw event reacts to mousedown instead of click events. This will cause the current solution to fail if the mouse button is hold longer than 300 ms. Another solution I thought of is to not do a request if the last point is within a certain pixel distance ...

simonseyock:
I tried drawing points, deleting points and right clicking on points and on the background.

geographika:
@simonseyock - many thanks for this - I've also tested all cases and it now works perfectly. 

I tried some debugging myself to resolve it but was trying to use `ol.events.condition.mouseActionButton(evt)` rather than `ol.events.condition.primaryAction(evt)` - I'll need to do some OL API docs reading!

PR #362, opened 2021-01-18
Set defaultClickEnabled based on toggleGroup state

geographika:
Relates to #351 - there are issues/race conditions when tools are destroyed (this seems to trigger the `onToggle` event. 
@simonseyock - sorry I hadn't seen you had a fix in a branch (https://github.com/simonseyock/cpsi-mapview/commit/98529209c80603b344a6109ef6d6bae7d3bb9ede) and had come up with a different resolution. 

It uses the `toggleGroup` property of the buttons - and listens to whenever this is fired. Any tools which can take the focus away from the default tool needs to have `toggleGroup: map` set. 

This makes it easier to track, and add new tools as you no longer have to worry about the `defaultClickEnabled` value in the tools themselves. 

Newly created tools, destroying tools etc. all seem to work fine with this approach.


geographika:
@simonseyock - I merged this to resolve the issue. might be worth a code review in a future sprint.

Issue #382, opened 2021-03-12
Abort Multiple Requests to the same WMS Map Layer


geographika:
If a user unchecks a grid filter, and then checks a new value then the same WMS layer is requested twice.
Ideally there would be a mechanism to abort any requests to the same layer so only the latest request is sent. 

https://github.com/compassinformatics/cpsi-mapview/blob/c9338fea016a382c46604ca2cb217c27bea2de9d/app/controller/grid/Grid.js#L199

This is not only applicable to grids but also when zooming in gradually, or with multiple pans in quick succession.
This has been raised at https://github.com/openlayers/openlayers/issues/8425 so this solution could possibly be added to all WMS layers in mapview by default.

simonseyock:
I introduced the proposed changes in https://github.com/compassinformatics/cpsi-mapview/pull/402

But I have to admit that I cannot reproduce the problem in the demo client. There are at most 3-4 concurrent requests that finish really quickly. What layers do you use to trigger this problem? Also I did not yet work with the grids and I don't know how to access them. Can you tell me?

geographika:
@simonseyock - WMS config sent by email. 
As grids use WFS they shouldn't be affected by this issue. A sample grid is available by right-clicking the Ruins layer:

![image](https://user-images.githubusercontent.com/490840/118480834-0de05780-b713-11eb-878d-064dad24c351.png)

simonseyock:
I have to confess that this layer also loads very fast for me. I see that the image source is overwritten at some points, but I do not see any cancelled requests. And after thinking about this I don't know if that is actually what you want :)
If the requests get cancelled they are already sent, and if the server does not support aborting a request (they rarely do, i highly doubt that a normal wms would do that) it would not get you any benefits in regard of server load.

What we can do though is a throttling that only requests wms once every second or something like that. Alternatively only request if no other request is following within a second or so.

geographika:
We had some performance issues with some WMS layers which was leading to a large queue of pending requests waiting in the browser. I was thinking cancelling the old ones would take some pressure off the server, but as you say once the request is triggered the server will do the work even if the client request is cancelled. 

[...]

PR #402, opened 2021-05-17
introduce image aborting

simonseyock:
Fixes: https://github.com/compassinformatics/cpsi-mapview/issues/382

Needs some testing.

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -273,20 +273,20 @@ Ext.define('CpsiMapview.factory.Layer', {
 
             if (layerConf.debounce !== undefined) {
                 var blankSrc = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=';
-                var img;
+                var currentImgElement;
                 var timeout;
                 olSourceConf.imageLoadFunction = function (image, src) {
-                    if (img) {
-                        img.src = blankSrc;
+                    if (currentImgElement) {
+                        currentImgElement.src = blankSrc;
                     }
-                    img = image.getImage();
+                    currentImgElement = image.getImage();
                     if (timeout) {
                         clearTimeout(timeout);
                     }
                     timeout = setTimeout(function () {
-                        img.src = src;
+                        currentImgElement.src = src;
                         timeout = undefined;
-                        img = undefined;
+                        currentImgElement = undefined;
                     }, layerConf.debounce);
                 };
             }
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -271,6 +271,16 @@ Ext.define('CpsiMapview.factory.Layer', {
 
         if (singleTile) {
 
+            const blankSrc = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=';
+            let img
+            olSourceConf.imageLoadFunction = function (image, src) {
+                if (img) {
+                    img.src = blankSrc;
+                }
+                img = image.getImage();
+                img.src = src;
+            }
+
             olLayerConf.source = new ol.source.ImageWMS(olSourceConf);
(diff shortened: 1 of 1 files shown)


simonseyock:
Usage in `default.json`:

```
// ...
 {
      "url": "some-url",
      "featureInfoWindow": true,
      "hasMetadata": true,
      "isBaseLayer": false,
      "openLayers": {
        "maxResolution": 1222.99245234375,
        "opacity": 0.9,
        "projection": "EPSG:3857",
        "visibility": false,
        "singleTile": true
      },
      "labelClassName": "labels",
      "layerKey": "SOME_WMS",
      "layerType": "wms",
      "serverOptions": {
        "layers": "some"
      },
      "text": "Works",
      "debounce": 2000 // <- value in ms
    }
// ...
```

app/factory/Layer.js
```diff
@@ -271,6 +271,26 @@ Ext.define('CpsiMapview.factory.Layer', {
...
 
+            if (layerConf.debounce !== undefined) {
+                var blankSrc = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=';
+                var img;
+                var timeout;
+                olSourceConf.imageLoadFunction = function (image, src) {
+                    if (img) {
+                        img.src = blankSrc;
+                    }
+                    img = image.getImage();
```
JakobMiksch, line 282: I find the names `img` and `image` a bit confusing, because on the first sight it is not clear to me what the difference is

JakobMiksch (commented):
looks fine to me, only one comment inline

geographika:
@simonseyock - thanks for this. It partially solves the issue. If I turn on 10 layers without this update and use the mouse wheel to zoom out 5 zoom scales then 50 WMS requests are made. Each of these is loaded (in batches of 6) until getting to the final WMS images that will be loaded into the map (so a delay for the user and database hits). 

With this patch the same action loads 10 WMS requests - after a wait of 2 seconds (using `"debounce": 2000` for all test layers), and then loads the almost identical 10 WMS requests after a 2-second pause. 
This saves the server 40 unnecessary requests (we can't use tiles for these layers as they are edited by the user so need to be dynamically generated). 

I'm not sure why the same requests are made twice. See image below along with timings. The only difference between the highlighted requests is the HEIGHT in the first is 786 and 789 in the second and a slightly modified BBOX. The images are almost identical. 

Is there anyway of avoiding the 2-second delay at the start? Could setting a `loading` flag be useful?

image

geographika:
Thinking more about it, from a user's perspective aborting the previous requests would mean the later ones don't have to wait for the queue to process. In the example above 50 requests would be made but 40 would be aborted so the browser would only be waiting for the last 10 to complete. 

The server taking the hit is less of a concern (as you say it would be hard with HTTP to know when a client aborts a request).

simonseyock:
Yes, I can change it to aborting. But you could also reduce the debounce time, maybe to something like 300 or 500 ms?

simonseyock:
I think debouncing is the usual approach in cases like this. Also I had the feeling that aborting image loads is quite clunky, when I was working with it. So I think debouncing is a more useful approach.

Another thing we could do is always wait for the last request until starting a new one. this way we always have just one request, but also start them as soon as possible.

simonseyock:
the problem with the two different request is something else that would need to get investigated. maybe there is a small change of the viewport size by extjs?

geographika:
Please leave as is - I'll experiment with some different settings. The map size must be changed somewhere to trigger the duplicate requests - will investigate further. 
I presume there is no particular advantage in using https://docs.sencha.com/extjs/6.7.0/classic/Ext.util.DelayedTask.html - does this serve a similar purpose to debouncing?

simonseyock:
it does exactly the same but saves 5 lines of code XD I added it.

geographika:
@simonseyock - I've been playing around with the settings - this is a very useful feature and will speed up the system (and save our database) for users who turn on all the layers and then use the mouse wheel for zooming!

One nice addition (if possible) would be to skip the timeout when turning on the layer (so change in visibility). There is no need to throttle/delay this action. Is this possible to add?

The small delay on panning/zooming once visible is hidden as OL redraws the existing images while waiting.

simonseyock:
added that.

geographika:
Thanks @simonseyock - please merge.

PR #403, opened 2021-05-17
Layer tree as child group

### LINKED ISSUE #352: Gazetteer Layers not appearing in the map
Relates to #70 

The layers are correctly added to the map at https://github.com/geoext/geoext3/blob/master/classic/form/field/GeocoderComboBox.js#L217 and when a feature is selected it is added to the layer https://github.com/geoext/geoext3/blob/master/classic/form/field/GeocoderComboBox.js#L370

However the layer itself is not visible (although its visible property is `true`). Adding the layer again to the map manually seems to fix this. The feature is therefore not highlighted, as it was previously in the image below. 

image

I believe it may be related to the loading of the layers into the layer tree at https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/LayerTreeController.js#L217

simonseyock:
This PR moves the layer group of the layer tree down one level. This allows for arbitrary layers to be added to the group without them showing up in the tree.

Additionally this makes the `displayInLayerSwitcher` property unnecessary.

Fixes #352

### DIFF
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -117,17 +117,6 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
     makeLayerStore: function () {
         var me = this;
 
-        // filter function for LayerTreeStore to hide unwanted layers in tree
-        var layerFilter = function (layerRec) {
-            var layer = layerRec.getOlLayer();
-            if (layer) {
-                // neither displayInLayerSwitcher=false (our flag) nor
-                // bp_displayInLayerSwitcher=false (flag of BasiGX)
-                return layer.get('displayInLayerSwitcher') !== false &&
-                    layer.get('bp_displayInLayerSwitcher') !== false;
-            }
-        };
-
         var treeJsonPromise = me.loadTreeStructure();
         treeJsonPromise.then(function (treeJson) {
             // save defaults for tree nodes from config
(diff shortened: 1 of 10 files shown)


geographika:
Sounds a good solution which simplifies things. 

Quick question - would any layers created and added to the map still remain hidden if the `displayInLayerSwitcher` filter is removed? For example in a derived application if a custom form runs the following with `addLayer` it remains hidden from the tree?

```
        var extraLayer= new ol.layer.Vector({
            source: new ol.source.Vector(),
            style: styleFunction,
            zIndex: 1000 // make sure the layer is always on top
        });

        map.addLayer(extraLayer);
```

simonseyock:
Yes, the tree would now be a child of the default layer group of the map. Only layers that are added to the layer group of the layertree (accessible via `map.get('layerTreeRoot')` if needed) are shown in the layer tree.

When you call `map.addLayer` the layer gets added to the main layer group of the map and is therefore not a child of the layer group of the layer tree.

JakobMiksch:
Looks fine for me  :+1:

simonseyock:
@geographika @JakobMiksch I just noticed that all layers were added to the map twice. Before that was not noticed because the layergroup was removed before adding them again.

I added a small change to remove them before adding them again. This could be done much cleaner if they would not be added here: https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/main/Map.js#L227
but instead passed over to the LayerController so they don't need to get fetched from the map here: https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/LayerTreeController.js#L261

I don't know how easy that would be and if that would cause other issues.

geographika:
So currently:

- `initComponent` in `CpsiMapview.view.main.Map` adds each of the layers in the config to the map, and then fires the `cmv-init-layersadded` event
- `CpsiMapview.controller.LayerTreeController` listens to the `cmv-init-layersadded` event and then constructs OL groups and moves the layers into them

Is the suggestion to add the layers in the `CpsiMapview.controller.LayerTreeController` tosimplify the process (and would need to be passed an array of layers)?
Or would the event simply pass the array of layers via the event and still add them?

The first option could cause issues if layers are expected to be available in the map `initComponent`. I don't see any issue with the second option though?

simonseyock:
I thought a little bit about that. I think the question is if there is always a layer tree. If that is the case the creation of the layers and the constructing of the group could be done in one step. If not it might just be fine the way it is now.

geographika:
@simonseyock - for all applications a layer tree is present. Reusing the config loading for a layertree-less application (e.g. a modern UI) would require lots of other changes.
So if you think it makes things simpler to move the loading to the layer tree it is fine with me.

geographika:
@simonseyock - I've tested on a couple of applications and all works perfectly. Will merge as-is and we can maybe revisit  the following in a separate pull request/issue. 

> creation of the layers and the constructing of the group in one step

PR #405, opened 2021-05-18
Allow clickevents on map through empty taskbar

### LINKED ISSUE #380: Taskbar Blocks Map Canvas
The taskbar currently extends across the width of the map at the bottom of the screen, even if there is only a single button. 
This prevents map tools, and the OL attribution button being used for that area of the map.
Ideally the taskbar would have a width corresponding to the size/number of buttons on display. 

image

JakobMiksch:
fixes #380 

Map can be controlled through empty taskbar.
The pointer-events on the taskbar are disabled via CSS and re-enabled on the buttons. 

Peek 2021-05-18 08-44

### DIFF
--- a/sass/src/view/main/Main.scss
+++ b/sass/src/view/main/Main.scss
@@ -15,6 +15,10 @@
         border-color: #778899;
         border-width: medium;
     }
+    // disable click events on the empty task bar
+    // this is necessary to controll the map
+    // and map interactions in this area
+    pointer-events: none;
 }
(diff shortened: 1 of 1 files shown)


simonseyock (approved):
looks good to me :+1:

geographika:
Excellent - thanks @JakobMiksch !

JakobMiksch:
@geographika yes, it is ready to merge

PR #407, opened 2021-05-19
Prevent multiple resolution change events

JakobMiksch:
This PR reduces the number of unnecessary fired events. 

Before during a zoom in the map, multiple (~ 5) `change:resolution` events were fired. This is a known behavior of OpenLayers (see https://github.com/openlayers/openlayers/issues/7402).

The fix is to use a `moveend` event. This, however, also fires on map-pan events. To prevent them, we check if the resolution has actually changed, otherwise we return.

Now, a zoom on the map fires only one event, like one would expect. This fix should be beneficial for the performance of the app.

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -152,7 +152,7 @@ Ext.define('CpsiMapview.factory.Layer', {
     },
 
     /**
-     * Creates an layer which renders shows a WMS for small scales and a WFS
+     * Creates a layer which renders a WMS for small scales and a WFS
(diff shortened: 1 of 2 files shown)


geographika:
Travis is failing with `const targetResolution`. Other than that looks good to me - thanks!

chrismayer (approved):
LGTM

geographika:
Looks good - please merge.

simonseyock (approved):

PR #408, opened 2021-05-20
Fix disappearing switchlayer and its text in layer tree

### LINKED ISSUE #390: Multiple Switch Layers not loading correctly
When two switchlayers are added layers are removed from the map and disassociated with the layer tree nodes. All works fine with a single switchlayer. 

I believe the following may be the key to finding the issue. Running `overlayCollection.getLength()` after the following call returns a count of one less than before - the first layer is removed by the `setAt` but the second layer is never added to the map. Trying to manually add the `newLayer` at this stage also fails to add the layer (but no warning or error):

https://github.com/compassinformatics/cpsi-mapview/blob/833ba027154bd6856f08eb097a0839eb7885c6ec/app/factory/Layer.js#L1021

I believe it may have something to do with the event handlers being triggered in `GeoExt.data.store.LayersTree` - `addLayerNode` etc. but suspending the events below caused similar issues with layers becoming disassociated with nodes.

```
        var treePanel = Ext.ComponentQuery.query('treepanel')[0];
        var treeStore = treePanel.getStore();
        treeStore.suspendCollectionEvents();
```

JakobMiksch:
fixes  #390 

This PR fixes two bugs of the switchlayer:

Scenario 1 "Switchlayer disappears":
- a switch layer is "last" in a group of the layer tree - this means it has the position `0` in the `ol.Collection`
- when the group is expanded, after switch of the layer, something removes this layer from its  `ol.Collection` but still keeps the reference in the treeStore
- replacing `setAt()` with `insertAt()/removeAt()` fixes the problem

Scenario 2 "missing text in layer tree":
- a switch layer is in a collapsed group
- the map resolution changes so that the switch layer changes
- its group in the layer tree is expanded
- then the name of the layer is not displayed anymore
- the problem is solved by re-adding the layer configuration its node in the treeStore

@geographika could you test this fix in one of your sample applications? So we can be sure it actually works in other settings as well. Further refactoring will be done in a follow-up PR

### DIFF
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -145,6 +145,14 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
             });
             me.getView().setStore(groupedLayerTreeStore);
 
+            // update possible switchlayers when collapsing their parent folder
+            // otherwise the node text would be wrong/empty
+            me.getView().getRootNode().on(
+                'beforeexpand',
+                me.updateSwitchLayerNodes,
+                me
+            );
+
             me.getView().getRootNode().cascade(function (node) {
                 // apply properties for tree node from corresponding tree-conf
                 if (node.getOlLayer()) {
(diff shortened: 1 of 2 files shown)


geographika:
@JakobMiksch - tested in one application with 2 switch layers in the same folder - all works fine - many thanks!

I think I ended up suspecting the following line in GeoExt but couldn't come up with a workaround:

https://github.com/geoext/geoext3/blob/f4c4b252d13e6737f4c4da9437eee09067132327/src/data/store/LayersTree.js#L320

geographika (approved):

chrismayer (approved):

chrismayer:
> I think I ended up suspecting the following line in GeoExt but couldn't come up with a workaround:
> https://github.com/geoext/geoext3/blob/f4c4b252d13e6737f4c4da9437eee09067132327/src/data/store/LayersTree.js#L320

@geographika Yes, somewhere at this place must be the root cause. We also could not find a quick solution therefore we came up with the fix in the CPSI-Mapview code of this PR.

chrismayer:
@geographika Is this good to go?

geographika:
@chrismayer @JakobMiksch - yes good to go - please merge!

PR #409, opened 2021-05-20
Refactor switchlayer

JakobMiksch:
This PR refactors the SwitchLayer functionality:
- the essential functionality has move to a new `SwitchLayer.js` util
- functions and constants have been extracted and documented
- some internal cleanup and logical grouping of code parts
- `SwitchLayer.spec.js` contains tests for the new util

The function `SwitchLayer.changeInternalLayer()` shares some code with `StyleSwitcherRadioGroup.onStyleChange()`, but there are still some substantial differences. Therefore, I am not sure if a potential reunion of some code would actually simplify things.

### DIFF
--- a/app/util/SwitchLayer.js
+++ b/app/util/SwitchLayer.js
@@ -1,0 +1,214 @@
+/**
+ * Util class for switchLayer related functions.
+ *
+ * @class CpsiMapview.util.SwitchLayer
+ */
+Ext.define('CpsiMapview.util.SwitchLayer', {
+    alternateClassName: 'SwitchLayerUtil',
+    singleton: true,
+
+    /** The possible states of the layer switch */
+    switchStates: {
+        ABOVE_SWITCH_RESOLUTION : 'cmv_above_switch_resolution',
+        BELOW_SWITCH_RESOLUTION : 'cmv_below_switch_resolution'
+    },
+
+    /**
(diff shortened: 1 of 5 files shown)


app/util/SwitchLayer.js
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Loops through all layers, identifies switch layers
+     * and replaces them if required
+     *
+     * @param {Number} resolution The new resolution
+     */
+    handleSwitchLayerOnResolutionChange: function(resolution) {
+        var layerCollection = BasiGX.util.Map.getMapComponent().getMap().getLayers();
+
+        this.checkSwitchLayersRecursively(layerCollection, resolution);
```
chrismayer, line 25: Please use `var staticMe = CpsiMapview.util.SwitchLayer;` instead of `this` like in the other functions.

chrismayer (approved):
LGTM, thanks! Just one minor comment.

geographika (approved):

geographika:
Thanks @JakobMiksch. Great to get this in its own class with tests. +1 from me on merging, and I can carry out further testing.

app/util/SwitchLayer.js
JakobMiksch: @chrismayer thanks for the hint, I fixed it

Issue #401, opened 2021-05-10
Configurable Preset Grid Filters


geographika:
Allow the `CpsiMapview.controller.grid.Grid` class to have configured pre-set filters. 
A new flag could also be set to clear all filters, or to reset to the pre-set filters. 

For simple filters, for example setting a year equal to a value the following works:
```js
    applyGridFilter: function (gridWindow, columnName, value) {

        var grid = gridWindow.down('grid');
        var column = grid.columnManager.getHeaderByDataIndex(columnName);
        if (!column.filter.value) {
            column.filter.value = value;
        }
    },

   // to call
   me.applyGridFilter(gridWindow, 'YearOfWorks', { eq: currentyear });
```

However list filters should also be supported which seems trickier as the filter menus need to be kept in synch:

```
        var menu = grid.columnManager.headerCt.getMenu();
        var menuitem = menu.items.get('columnItem');
        var columnMenu = menuitem.getMenu();
```

The current demo does not currently have any list filters, so this would need to be setup for testing. 

![image](https://user-images.githubusercontent.com/490840/117658673-4cb66080-b19b-11eb-90ed-cb82eb835a34.png)

The grid filters should be configurable from the [config file](https://github.com/compassinformatics/cpsi-mapview/blob/master/resources/data/layers/default.json)
The JSON config objects could match the [Ext.util.Filter](https://docs.sencha.com/extjs/6.7.0/classic/Ext.util.Filter.html) structures as much as possible:

```json

"gridFilters": [{
    "property": "age",
    "value": 42,
    "operator": "<"
}]
```

Unit tests would also be useful for this new feature.

JakobMiksch:
@geographika do you have layer which I can use for testing the `list` filter?
I tried `tourist_info`, but in contrast to `ruins` CPSI cannot seem to display it as grid. This is the error from the Server:
```
<?xml version="1.0" encoding="UTF-8"?>
<ows:ExceptionReport xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ows="http://www.opengis.net/ows/1.1" version="2.0.0" xml:lang="en-US" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd">
  <ows:Exception exceptionCode="OperationProcessingFailed" locator="mapserv">
[...]

PR #410, opened 2021-05-21
Configurable preset filters

JakobMiksch:
references #401

This PR is already fully functional for `string` and `number` based filters, but is missing the `list` filter preset.

example configuration:
```json
[...]
      "gridFilters": [
        {
          "property": "code",
          "value": "42",
          "operator": ">"
        },
        {
          "property": "name",
          "value": "old",
          "operator": "="
        }
      ],
[...]
```

With the button ":star: Apply Preset Filters", the preset filters can be applied to the grid. See screenshot below.

- This works so far for `number` and `string` filter. 
- basic test classes have been created

**remaining:**
- [x] the `list` filter needs to be implemented
- [ ] `boolean` filter
- [ ] `date` filter

image

### DIFF
--- a/app/controller/grid/Grid.js
+++ b/app/controller/grid/Grid.js
@@ -735,8 +735,8 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
             var operator = filterDef.operator;
 
             if (!columnName || !value || !operator) {
-                console.warn('Preset filter is not properly defined.');
-                console.warn(filterDef);
+                Ext.log.warn('Preset filter is not properly defined.');
+                Ext.log.warn(filterDef);
                 return;
             }
 
--- a/app/controller/grid/Grid.js
+++ b/app/controller/grid/Grid.js
@@ -16,10 +16,10 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
     ],
 
     /**
-    * The currently active spatial filter for the layer.
-    *
-    * @cfg {Ext.util.Filter} spatialFilter
-    */
+     * The currently active spatial filter for the layer.
+     *
+     * @cfg {Ext.util.Filter} spatialFilter
+     */
     spatialFilter: null,
 
     /**
@@ -276,10 +276,10 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
     },
 
     /**
-    * Hide the loading mask when the store has loaded
-    *
-    * @private
-    */
+     * Hide the loading mask when the store has loaded
+     *
+     * @private
+     */
     onWfsStoreAfterLoad: function (store, features, success) {
 
         var grid = this.getView();
@@ -409,13 +409,13 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
         }
     },
     /**
-    * Enable and disable paging for the grid.
-    * Disabling paging allows all records to be loaded into the
-    * grid for an Excel export. Enabling paging improves load
-    * performance.
-    *
-    * @private
-    */
+     * Enable and disable paging for the grid.
+     * Disabling paging allows all records to be loaded into the
+     * grid for an Excel export. Enabling paging improves load
+     * performance.
+     *
+     * @private
+     */
     togglePaging: function (checkBox, checked) {
 
         var me = this;
@@ -449,10 +449,10 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
     },
 
     /**
-    * Export the current records in the grid to Excel
-    *
-    * @private
-    */
+     * Export the current records in the grid to Excel
+     *
+     * @private
+     */
     exportToExcel: function () {
 
         var me = this;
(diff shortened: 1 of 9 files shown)


app/controller/grid/Grid.js
```diff
@@ -688,5 +686,120 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
...
+            if (!filterDef || !Ext.isObject(filterDef)) {
+                return;
+            }
+
+            var columnName = filterDef.property;
+            var value = filterDef.value;
+            var operator = filterDef.operator;
+
+            if (!columnName || !value || !operator) {
+                console.warn('Preset filter is not properly defined.');
```
geographika, line 738: Use `Ext.log.warn` rather than `console.log`?

app/view/grid/Grid.js
```diff
@@ -110,7 +110,14 @@ Ext.define('CpsiMapview.view.grid.Grid', {
             text: 'Clear Sorting',
             glyph: 'f039@FontAwesome',
             handler: 'onClearSort'
-        }, '->',
+        },
+        {
+            xtype: 'button',
```
geographika, line 115: Have this button hidden by default, but display for the ExampleGrid? Otherwise all existing grids that inherit from this class get the button.

app/controller/grid/Grid.js
```diff
@@ -688,5 +686,120 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
         };
 
         viewModel.setStores(stores);
+    },
+
+    /**
+     * Applies preset filters from the configuration
+     * to the grid.
+     */
+    onApplyPresetFilters: function () {
```
geographika, line 695: Call this automatically when the grid is created?

geographika:
@JakobMiksch - I have added a new layer and associated grid that allows the list testing in #413

image

I think I ran into a few issues trying to ensure that the filter checkboxes (in screenshot above) were updated as well as the grid data.

geographika:
@JakobMiksch - in answer to your question - it would be useful to get all filters into place, including boolean and date. However for current applications these are not yet required so can be a lower priority. 

One UI issue to consider is how best to carry out the following actions:

- Clear Filters (as currently)
- Reset Filters (to remove all filters except for the defaults)

image

I was thinking maybe a [split Button](https://examples.sencha.com/extjs/7.4.0/examples/kitchensink/?classic#split-buttons) could be the best approach - with "Reset Filters" being the default option, but also allowing a user to "Clear Filters". 

The split button would only be necessary if preset filters applied to a grid, otherwise it could revert to a standard "Clear Filters" button. 

Thoughts/suggestions on alternatives welcome.

app/controller/grid/Grid.js
JakobMiksch: done with 3fe969e

app/view/grid/Grid.js
JakobMiksch: Good point. I changed it so the button will only shown if the layer has the property "gridFilter". I will push the commits later

app/controller/grid/Grid.js
JakobMiksch: yes, we can do that. Shall the preset filter always be applied directly? Or shall we have a config for that?

app/view/grid/Grid.js
JakobMiksch: here is the respective commit c8e8c23
@geographika please let me know if we can close this conversation

JakobMiksch:
@geographika The latest commits address your comments and also enables the `list` filter
`boolean` and `date` are not implemented yet, since I cannot test them on an existing example grid. I tried mocking a column, but it did not work out yet.

geographika:
Thanks @JakobMiksch - all filters working well. I'm happy to merge if you are. 

There is an additional change it would be good to get your thoughts on (and I probably should have added it to the original spec). 

When the application is first loaded the layer associated with the grid is unfiltered:

image

It is automatically filtered when the grid is opened:

image

What would be the best approach to have the layer filtered when the app is loaded? Would a grid need to be created in the background? Or could the preset filters in the config be applied to the layer directly?

If there is time left in the sprint I can detail this in a separate issue.

JakobMiksch:
@geographika I expect less troubles setting the preset filters via the grid, because then the sync with the UI might be easier. I did a quick prototype using the grid. I could already set the preset filters,but there are still some issue. I would be fine to move this to a separate issue.

app/controller/grid/Grid.js
```diff
@@ -688,5 +703,151 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
...
+            if (!column || !column.filter || !column.filter.type) {
+                return;
+            }
+            var columnType = column.filter.type;
+
+            switch (columnType) {
+                case 'string':
+                    // only equal is supported for string
+                    if (operator !== '=') {
+                        return;
```
simonseyock, line 793: Would it make sense to print a warning here?

app/controller/grid/Grid.js
```diff
@@ -688,5 +703,151 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
...
+                    }
+                    column.filter.setValue(value);
+                    break;
+                case 'number':
+                    var filterValue = me.createNumberFilterValue(
+                        operator,
+                        value
+                    );
+                    if (!filterValue) {
+                        return;
```
simonseyock, line 803: Would it make sense to print a warning here?

simonseyock (approved):
Looks very nice! Just some minor remarks.

app/controller/grid/Grid.js
JakobMiksch: good point. I added a warning with 
77c7639

app/controller/grid/Grid.js
JakobMiksch: Good point. I added a warning with 77c7639

PR #412, opened 2021-05-27
Update ol

simonseyock:
This PR uses branches on geoext and BasiGX to use the updated openlayers version.

@geographika it would be nice if you could test this.

To do so, you have to run `npm i` in the `cpsi-mapview` directory and update the submodules via `git submodule update` (should be the right command).

Some more work on the geoext and BasiGX repositories needs to be done to consider this finished, apart from any bugs that might be encountered.

The branches are:
https://github.com/simonseyock/geoext3/tree/update-ol
https://github.com/terrestris/BasiGX/tree/update-ol


simonseyock:
There might be a small complication at the moment as I rebased the `update-ol` branch on geoext right now. If there are any complications just checkout out the linked `update-ol` branch. It will soon get merged into the master then I will update this PR.

simonseyock:
I updated the BasiGX branch and this to use the current geoext master with ol 6.

simonseyock:
I forgot to commit the `app.json` :sweat_smile:

geographika:
@simonseyock - excellent work progressing all of these updates - thanks.

I've tested the branch locally - both is cpsi-mapview and in a more complex application. I've updated the demo recently to mimic more how the various cpsi-mapview features are used in a readl-world application. 

It is hard to know where to focus testing, but I've noticed a couple of things.

1. With the change at https://github.com/geoext/geoext/pull/686 requiring all layers to have `ol.collection` - I presume this means all layers created in cpsi-mapview need to be created using the new approach?

```js
var layer = new ol.layer.Vector({
                source: new ol.source.Vector({
                    features: new ol.Collection([
                        feature
                    ])
                })
            });
```

For example in the digitising tools, which are currently broken (with the error `"GeoExt.data.store.Features needs to be configured with a feature collection or with a layer with a source with a feature collection."`

https://github.com/compassinformatics/cpsi-mapview/blob/6e297c19e59ec8d5042fc3dff8faf2f19fc8e7dd/app/controller/button/DigitizeButtonController.js#L324

2. When using the grid "Select by Shape" I get `DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The canvas has been tainted by cross-origin data.`  - although these errors are caught in ol so may not be an issue (line 264 in `Layer.js`: `data = newContext.getImageData(0, 0, 1, 1).data;`

It will probably be easier to test if this is rebased onto master where the demo has been cleaned-up.

simonseyock:
1. yeah if the layer is connected to a features store this has to be done. I missed the digitize button somehow. I introduced that error message so the cases can be found quickly.
2. That seems familiar. I think this is a problem that occurs for `.getFeaturesAtPixel` if a layer with cross-origin problems is part of the currently drawn map.

simonseyock:
I updated the `DigitizeButtonController` and hope this was the only class i oversaw :D

simonseyock:
I have rebased this PR to the current master. @geographika It would be nice if you could check it out again. The easiest way to do that after the rebase is to delete the branch locally (`git branch -D update-ol`) and then check it out again (`git fetch --all; git checkout origin/master; git checkout -b update-ol`. There are more elegant ways somewhere out there but this should work :)

simonseyock:
I tried to reproduce the CORS problem with "Select By Shape", but for me it just works. Or are you saying that the error is happening inside of ol? If that is the case, then I think we can savely ignore it.

geographika:
>  Or are you saying that the error is happening inside of ol? If that is the case, then I think we can savely ignore it.

Yes - it was only apparent because I was running in the debugger - it is simply caught and ignored, so as you say not an issue.

geographika:
@simonseyock - I've been testing the branch/OL6 on a more complex project using cpsi-mapview and done some debugging. I've some questions!

1. I noticed all the OL events no longer take a scope:

```js
//  me.drawInteraction.on('drawend', me.handleLocalDrawEnd, me); // me is removed
me.drawInteraction.on('drawend', me.handleLocalDrawEnd);
```

In the DigitizeButtonController.js there is a custom event fired:

```js
resultSource.dispatchEvent({ type: 'featuresupdated', modifications: modifications });
```

In the client application I have a model that listens to this event:

```js
var me = this;
edgesLayer.getSource().on('featuresupdated', me.onGeometryUpdate, me); // need to change scope here, but can no longer use me
```

How would I pass in the scope to the `me.onGeometryUpdate` function? By default scope is set back to the layer source. 

2. I'm getting `f is undefined` at: https://github.com/compassinformatics/cpsi-mapview/blob/ef7a3316df00941122cdb9c932a5fd03bd1b42c5/app/controller/button/DigitizeButtonController.js#L952

These are not `undefined` in the `allFeatures` array before the loop. I tried reversing the loop but still get the same error. Any ideas what could be different here? Could it be related to the 2-way synching?

These are the only 2 issues I've encountered so far.

simonseyock:
For point 1: I posted some upgrade notes here: https://github.com/geoext/geoext/releases/tag/v4.0.0
There is a specific example how to solve your case.
For point 2: I just solved this. The problem is that `getFeatures()` now returns the same array that is manipulated with `removeFeature`. A simple `.slice(0)` did the trick.

geographika:
Sorry I did read the release notes and then forgot about them when I hit the first issue. 
Both issues I'm aware of are fixed so +1 for merging.

simonseyock:
I actually reworked them yesterday and added the examples. So it might not have been that obvious. I will merge this then now.

simonseyock:
@geographika I have written a more thorough walkthrough to update an application: https://github.com/terrestris/BasiGX/wiki/Update-application-to-ol-6.5.0,-geoext-4.0.0,-BasiGX-3.0.0

PR #423, opened 2021-06-08
Add timeout to first show of tooltip

### LINKED ISSUE #224: Tooltips Popup Cut at Bottom
The first time a tooltip is displayed the bottom is cut off (as shown below). The next time it is displayed this is fixed. 

image

simonseyock:
This fixes #224

Somehow extjs seems to have a problem with the very first render. It renders it correctly at first, but a fraction of second later the design changes a tiny bit and the calculated width is no longer big enough. This might as well be a problem with the css eniginge or some obscure css sideeffect that I was not able to find.

I investigated several solutions and debugged the extjs code a little bit, and I abstained from using `width: auto !important` and instead settled for a timeout of 200ms for the first render.

### DIFF
--- a/app/view/layer/ToolTip.js
+++ b/app/view/layer/ToolTip.js
@@ -62,24 +62,17 @@ Ext.define('CpsiMapview.view.layer.ToolTip', {
         }
     },
 
-    /**
-     * @private
-     */
-    initComponent: function () {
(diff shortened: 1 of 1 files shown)


geographika:
Thanks for the fix! Please merge if you are happy.

Issue #417, opened 2021-06-03
Add Reset Filters context-menu option


geographika:
Add an option on the context-menu for a layer to `Reset Filters`. This would include both attribute and spatial filters set in the grid, and would likely need to call:

https://github.com/compassinformatics/cpsi-mapview/blob/6e297c19e59ec8d5042fc3dff8faf2f19fc8e7dd/app/controller/grid/Grid.js#L588

If there are no filters then the menuitem should be disabled. If there is no associated grid it should not appear at all. 

With the introduction of #410 it is also possible for a layer/grid to have default preset filters. If this is the case then rather than removing all filters it would revert to the preset filters (so probably simplest to add this logic to the `clearFilters` function linked to above). If a user wants to see all data they can still manually remove all filters from the grid. 

The option would be created using the same approach as for other layer context menus e.g. https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/menuitem/LayerRefresh.js

![image](https://user-images.githubusercontent.com/490840/120645390-aea37680-c478-11eb-99d4-c9481bbe127b.png)

JakobMiksch:
@geographika a question:
when there is a layer with a grid and preset filters. Should the button "Reset Filters" only be shown, when the preset filters are currently **not** applied?

geographika:
I think if there are preset filters it should always be shown. If it is easy to disable the menuitem when only preset filters are applied that would be good to have but not essential.

PR #428, opened 2021-06-09
Add "Reset Filter" menu item on layer tree

JakobMiksch:
references #417 

This PR adds the menu item "Reset Filter".

image

It removes all filters by calling the [`clearFilters`](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/grid/Grid.js#L594-L612) method of the layer's grid. Additionally, if the layer has preset filters, they will be applied with the [`applyPresetFilters`](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/controller/grid/Grid.js#L731-L735) method afterwards.

The menu item is only shown for layers that have a grid. And additionally if one of these conditions are fulfilled:
- the layer-config contains "preset filters"
OR
- the layer is currently filtered

For checking if the layer is currently filtered it is necessary to create the grid earlier. That's why the creation of the grid in the class [`LayerGrid`](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/menuitem/LayerGrid.js) is done as soon the context menu of a layer is opened. I could not feeld any performance limitation with this approach.

~~The functionality does not work for the `ruins` layer. Probably because its synchronization between grid and layer is currently broken.~~

### DIFF
--- a/app/controller/grid/Grid.js
+++ b/app/controller/grid/Grid.js
@@ -544,10 +544,29 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
         var grid = me.getView();
         var store = grid.getStore();
 
-        if (store.isEmptyStore !== true) {
-            var layer = store.getLayer();
+        var layer = me.getOlLayer();
+        if (store.isEmptyStore !== true && layer) {
             layer.setVisible(show);
         }
+    },
+
+    /**
+     * Returns the layer of the grid.
+     *
+     * @returns {ol.layer.Base} The grid's layer
+     */
+    getOlLayer: function() {
+        var me = this;
+        var viewModel = me.getViewModel();
+        var wmsLayerKey = viewModel.get('wmsLayerKey');
+        var vectorLayerKey = viewModel.get('vectorLayerKey');
+        var layer;
+        if (wmsLayerKey) {
+            layer = me.getLayerByKey(wmsLayerKey);
+        } else if (vectorLayerKey) {
+            layer = me.getLayerByKey(vectorLayerKey);
+        }
+        return layer;
     },
 
     /**
@@ -564,14 +583,6 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
      */
     onShow: function () {
         this.toggleLayerVisibility(true);
-    },
-
-    /**
-     * Apply preset filters once grid is created
-     */
-    onBoxReady: function () {
-        var me = this;
-        me.onApplyPresetFilters();
     },
 
     /**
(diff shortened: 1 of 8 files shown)


JakobMiksch:
I have tested it with the ruins layer  as well. The functionality works for individual layer, but there is still interference between the filters of both layers. I am currently looking for a solution

JakobMiksch:
I added an update with a workaround for the interfering layers

JakobMiksch:
@geographika this PR can be reviewed again. It works with two layers. The interference between them is solved with the workaround (`grid.show()` / `grid.hide()`).
The presets are applied automatically applied to grid once it is opened by the user.

chrismayer:
@JakobMiksch do we still have to create the grid windows after layer creation when following the approach "apply of predefined filters once user opens the grid"?

JakobMiksch:
@chrismayer we have to create it before the item "Reset Filters" is clicked. Before it was created once the user opened the grid via the respective menu item. Alternatively we can create the grid once the user right clicks on the layer. This would less created grids at the startup. 
We can also do the grid creation at this point.

chrismayer:
I'd prefer to create the grid windows when needed and not all in advance since this does not scale very well (and is very ugly :wink: ).
I suggest doing the following (little) rewrite: Separate the function to create the grid window and call it when either the menu item for ``Open Data Grid`  or `Reste Filter` is created. This would IMHO simplify things a lot ...

geographika:
The approach suggested by @chrismayer sounds good (creating the grids when the context menu is opened). The client apps make extensive use of grids (up to 70), so only creating these when necessary is important.

JakobMiksch:
@geographika @chrismayer here my updated changes:

- the grid creation has moved to a separate util which will only be called once the grid is required. This is a way cleaner approach than before. A basic test for this util is provided.
- the grid had to be created with a **new** empty store (see [here](https://github.com/compassinformatics/cpsi-mapview/pull/428/files#diff-d4bd710b2f42fc0d10f8eccf94ad29249f3558642734e79a13d5163fe03820c2R64-R71)). Otherwise Ext would use a default store. We suppose that Ext used the same default store for all grids and therefore the filters we applied to both under some circumstances. 
- Bug fix: the "reset filters" menu item was not shown if the layer (1) did not have preset filters AND (2) it was spatially filtered. This is because the spatial filters are not retrievable through the store, but need the be retrieved via the grid's controller
- Bug fix: the method `store.getLayer()` did not work anymore. I replaced it with the new method `getOlLayer()` which I reused in other places as well.
- Bug fix: the grid window had a function that on close, the layer should be disabled. This function did not work in the last time, and just started to work again after introducing `getOlLayer()`. Apparently this behavior was not noticed. Hence, I removed the respective code. To sum it up: Now, when the grid is closed, the layer stays visible. If the opposite is desired, I can of course change this.

(hints for testing:  I temporarily removed `maxScale` and the `gridFilters`  from "ruins")

JakobMiksch:
just rebase this branch to the latest `master`

app/controller/grid/Grid.js
```diff
@@ -827,7 +812,9 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
                 default:
                     break;
             }
-        });
+        }
+
+        );
```
simonseyock, line 817:
```suggestion
        });
```

simonseyock (approved):
Looks very good in my opinion!

geographika (approved):

geographika:
Thanks @JakobMiksch for this and the design notes. I had a quick test and all worked well with the branch. +1 on merging from me.

PR #429, opened 2021-06-09
Demo updates

geographika:
- Simplifies tree structure
- Changes ruin layer to `noCluster` to allow grid synch (see #428 notes)
- Set `passThroughFilter` to `true` by default for stores


JakobMiksch:
@geographika looks fine for me

geographika:
@JakobMiksch - seems to be an issue with clustered vector layers not refreshing. I'll create an issue for that as a reminder for a future sprint, but ruins layer/grid should be in sync again.

PR #431, opened 2021-06-10
disable draw while on edge

### LINKED ISSUE #421: Sync map feature selections with grid selections
The `CpsiMapview.controller.button.DigitizeButtonController` based tools allow selection of records in a grid which selects features in the map. In the screenshot below the second grid record is selected, and it is automatically highlighted in red on the map. 

image

This is handled with:

https://github.com/compassinformatics/cpsi-mapview/blob/6e297c19e59ec8d5042fc3dff8faf2f19fc8e7dd/app/controller/button/DigitizeButtonController.js#L1097

It would be useful to also be able to select features in the map to update the selection in the grid. This would likely require the layer to have onClick events, which would prevent the digitising tool being triggered again. 

If there is an approach that would allow a generic mixin to be added to automatically sync grid and layer selections to GeoExt then this would be ideal (similar to the store/layer synching in https://github.com/geoext/geoext/pull/683). 

For testing the point digitising tool can be used to select roads in Tipperary - https://compassinformatics.github.io/cpsi-mapview/#map=19/-877886.9/6894435.27/0

simonseyock:
This PR disables draw when clicking on an edge for the digitize tool. This is needed if selection of digitized features in the map should be possible.

It also adds a selection model that allows map selection to the digitize button example.

https://github.com/geoext/geoext/pull/690 should be added for a better selection and a bugfix.

Fixes #421

### DIFF
--- a/app/controller/button/DigitizeButtonController.js
+++ b/app/controller/button/DigitizeButtonController.js
@@ -54,7 +54,12 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
     /**
      * OpenLayers snap interaction for better vertex selection
      */
-    snapInteraction: null,
+    snapVertexInteraction: null,
+
+    /**
+     * OpenLayers snap interaction for better edge selection
+     */
+    snapEdgeInteraction: null,
 
     /**
      * CircleToolbar that will be set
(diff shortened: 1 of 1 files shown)


app/controller/button/DigitizeButtonController.js
```diff
@@ -341,11 +349,17 @@ Ext.define('CpsiMapview.controller.button.DigitizeButtonController', {
             }
         }
 
+        me.setDrawInteraction(me.drawLayer);
+        me.setModifyInteraction(me.drawLayer);
+        me.setPointerInteraction();
```
JakobMiksch, line 354: is there intentionally no argument?

JakobMiksch (commented):
@simonseyock the code looks clean and alright. One question inline.

app/controller/button/DigitizeButtonController.js
simonseyock: the pointer interaction does not have a layer filter, but interacts with the `drawLayer` directly so it is not needed to pass it here.

JakobMiksch: alright

geographika (approved):

PR #432, opened 2021-06-10
IE 11 fixes

simonseyock:
While investigating the ie11 bug I found out that the openlayers 6.5.0 version we used was not working in ie11. I created a new version that is packaged in `@geoext/openlayers-legacy@6.5.4`.

**Important**: Only the `ol.js` file will work in IE11. The `ol-debug.js` will not work!

### DIFF
--- a/app.json
+++ b/app.json
@@ -208,7 +208,7 @@
      */
   "js": [
     {
-      "path": "./node_modules/@geoext/openlayers-legacy/dist/ol-debug.js"
+      "path": "./node_modules/@geoext/openlayers-legacy/dist/ol.js"
     },
     {
       "path": "./node_modules/geostyler-sld-parser/browser/sldStyleParser.js"
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "homepage": "https://github.com/compassinformatics/cpsi-mapview#readme",
   "dependencies": {
-    "@geoext/openlayers-legacy": "^6.5.3",
+    "@geoext/openlayers-legacy": "^6.5.4",
     "eslint": "^6.2.1",
     "font-gis": "^1.0.2",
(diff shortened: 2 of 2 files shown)


simonseyock:
There is one more problem: geostyler-openlayers-parse does not seem to be ie 11 compatible. I will ask if this is intended. If not we can try to throw it into babel and see what comes out.

geographika:
For other IE11 fixes I've added in polyfills using https://polyfill.io/v3/url-builder/ for any missing functions. Any idea what is causing it to break?

simonseyock:
the geostyler build does not seem to transpile arrow functions down to `es5`

geographika:
Would taking the source and pasting it into https://babeljs.io/repl, and then saving the output in the `/lib` folder be an option?

simonseyock:
That would be possible, but I would prefer if we first check if geostyler is meant to not run in ie 11 and if that is the case, add a `postinstall` step that transpiles geostyler via babel automatically. That is the same effort and more future proof.

chrismayer:
As far as I know GeoStyler is meant to be running in IE 9+ but I am not sure if this has full attention (and surely no focus) in the GeoSyter development process.

` "not ie < 9"`source https://github.com/geostyler/geostyler/blob/master/package.json#L32

simonseyock:
Good to know, than I will try to fix this upstream if that's fine with you @geographika

geographika:
> 
> 
> Good to know, than I will try to fix this upstream if that's fine with you @geographika

Even better!

geographika (approved):

simonseyock:
It should be fixed with this: https://github.com/geostyler/geostyler-openlayers-parser/pull/330

simonseyock:
This way geostyler seems to work in IE 11.

Issue #427, opened 2021-06-09
StreetView tool missing icon


geographika:
The icon (in the image below) for the StreetView tool no longer displays. See https://compassinformatics.github.io/cpsi-mapview demo. 

![image](https://user-images.githubusercontent.com/490840/121348896-32ea7380-c929-11eb-865a-0928a868e10c.png)

There no longer seems to be a need for the SV layer to be added to a group using the code at: https://github.com/compassinformatics/cpsi-mapview/blob/3edacca23c9e2d2c1e3e24b84959e66868917308/app/controller/button/StreetViewTool.js#L137

However even when switching to a simple `me.map.addLayer(me.vectorLayer)` the icon is still not visible. The layer seems to be present in the map and visible. There doesn't appear to be any issue with the icon image path or deployment. Changing the zIndex of the layer doesn't make the icon appear.

geographika:
Closed with #436

PR #436, opened 2021-06-10
Simplify adding the StreetView to the map

geographika:
With the switch to putting the tree in a group layer in #403, the code to add the StreetView layer can be simplified. It can be hidden from the layer tree like other tool layers (measure, gazetteers etc.). 

However as per #427 the icon is no longer appearing when clicking on the map. This was also the case prior to any changes in this pull request. 

I'm out of ideas why this is no longer appearing. It seems to be added to the (correct) map, and visible.


geographika:
It seems as though the cause of the icon disappering was the `vectorSource.refresh();` call. Removing it and the icon displays correctly again (and is still in synch with moving within the SV window).

PR #437, opened 2021-06-14
Minimize all

### LINKED ISSUE #420: Add a Minimize All button
Add a button that can minimize all [Cpsi Windows](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/window/MinimizableWindow.js) to the [Cpsi Toolbar](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/view/toolbar/MinimizedWindows.js) with a single-click. This was available in the old version of MapView with the following code:

```js
            var minAllButton = Ext.create(Ext.button.Button, {
                itemId: 'minimizeAllButton',
                scope: this,
                handler: function (button, event) {
                    Ext.WindowManager.each(function (w) {
                        if (w.minimize && w.isVisible()) {
                            w.minimize();
                        }
                    });
                },
                tooltip: 'Minimize all windows', // TOFIX tooltips keep redrawing blocking button clicks
                tooltipType: 'title',
                iconCls: 'icon-minimizeAll'
            });
            return minAllButton;
```

The button could be placed on the bottom toolbar in the demo app.

simonseyock:
This adds a new `cmv_minimize_all_button`. Somehow the FontAwesome glyph does not work. I tested it with different glyphs that worked, but https://fontawesome.com/v4.7/icon/window-minimize did not. I tried to use the code that was listed on the page under `unicode` maybe that was wrong. Do you know more @geographika ?

Fixes #420

### DIFF
--- a/app/controller/button/MinimizeAllButtonController.js
+++ b/app/controller/button/MinimizeAllButtonController.js
@@ -1,5 +1,5 @@
 /**
- * This class is the controller for the button 'MinimizeAllButton'
+ * This class is the controller for the button 'MinimizeAllButton'.
  */
 Ext.define('CpsiMapview.controller.button.MinimizeAllButtonController', {
     extend: 'Ext.app.ViewController',
--- a/app/controller/button/MinimizeAllButtonController.js
+++ b/app/controller/button/MinimizeAllButtonController.js
@@ -1,0 +1,17 @@
+/**
+ * This class is the controller for the button 'MinimizeAllButton'
+ */
+Ext.define('CpsiMapview.controller.button.MinimizeAllButtonController', {
+    extend: 'Ext.app.ViewController',
+
+    alias: 'controller.cmv_minimize_all_button',
+
+    onClick: function () {
+        var windows = Ext.ComponentQuery.query('cmv_minimizable_window');
+        Ext.Array.forEach(windows, function (win) {
+            if (win.isVisible()) {
+                win.minimize();
+            }
+        });
+    }
+});
--- a/app/view/button/MinimizeAllButton.js
+++ b/app/view/button/MinimizeAllButton.js
@@ -16,7 +16,7 @@ Ext.define('CpsiMapview.view.button.MinimizeAllButton', {
     /**
      * The icon the button should use
      */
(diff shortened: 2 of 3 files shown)


app/controller/button/MinimizeAllButtonController.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * This class is the controller for the button 'MinimizeAllButton'
```
JakobMiksch, line 2:
```suggestion
 * This class is the controller for the button 'MinimizeAllButton'.
```

app/view/button/MinimizeAllButton.js
```diff
@@ -0,0 +1,38 @@
...
+    xtype: 'cmv_minimize_all_button',
+
+    requires: [
+        'CpsiMapview.controller.button.MinimizeAllButtonController'
+    ],
+
+    tooltip: 'Minimize all windows',
+
+    /**
+     * The icon the button should use.
```
JakobMiksch, line 17:
```suggestion
     * The icon the button should use
```

JakobMiksch (commented):
Looks good. Minor formatting notes.

simonseyock:
Thanks for the review!

geographika:
> 
> 
> This adds a new `cmv_minimize_all_button`. Somehow the FontAwesome glyph does not work. I tested it with different glyphs that worked, but https://fontawesome.com/v4.7/icon/window-minimize did not. I tried to use the code that was listed on the page under `unicode` maybe that was wrong. Do you know more @geographika ?
> 

Works great - thanks @simonseyock!

It looks like ExtJS 6.x uses FontAwesome 4.4.0 according to [the docs](https://docs.sencha.com/extjs/6.7.0/guides/core_concepts/font_ext.html#core_concepts-_-font_ext_-_available_font_packages). 

Looking at https://www.fontawesomecheatsheet.com/font-awesome-cheatsheet-4x/ - it appears the minimize glyph was added in 4.7. 

Maybe go with `glyph: 'f068@FontAwesome'` (a minus sign) instead?

geographika:
Unless there are any objections I'll merge this later today. Thanks @simonseyock !

PR #466, opened 2021-08-03
Add a new style for clustered layers

geographika:
This checks if the layer source is clustered, and if it is then it wraps the style function generated from the SLD, and adds a new clustered style for points. 

image

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -890,8 +890,6 @@ Ext.define('CpsiMapview.factory.Layer', {
      */
     loadSld: function (mapLayer, sldUrl) {
 
-        var me = this;
-
         Ext.Ajax.request({
             url: sldUrl,
             method: 'GET',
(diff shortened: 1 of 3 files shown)


app/factory/Layer.js
```diff
@@ -880,6 +880,30 @@ Ext.define('CpsiMapview.factory.Layer', {
         return olSourceProps;
     },
 
+    /**
+     * Create a point style for clustered features
+     * @param {any} featCount the number of features in the clustered feature
+     */
+    createClusterStyle: function (featCount) {
```
chrismayer, line 887: Wouldn't this be placed better in `CpsiMapview.util.Style`?

chrismayer (approved):
LGTM, just one suggestion to move the `createClusterStyle` to `CpsiMapview.util.Style`, which I leave up to you.

app/factory/Layer.js
geographika: Thanks for the review @chrismayer - yes it would make more sense to move this to the style util class

PR #479, opened 2021-08-24
Rename functions and classes for clarity and add docstrings

geographika:
This pull request addresses some of the issues highlighted by @chrismayer in a code review.
The following are addressed:

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/form/ControllerMixin.js#L40
Maybe fire also an event like 'savefailed', like it is done for success case.

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/form/RightInfoField.js#L2
API-Docs are maybe wrong here? I think this is no Mixin, is it?

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/form/ValidationMessages.js#L7
I'd suggest to rename it with a 'Mixin' suffix in the class name to be consistent.

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/form/ViewModelMixin.js#L26
The 'destroyUtil' naming seems non intuitive here

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/form/ViewModelMixin.js#L43
Missing API-docs should be added. Also applies for other Mixin classes.

There are still plenty of doc updates to add in and also the types issue below to address, so this pull request is hopefully the first of several (time permitting). 

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/form/ViewModelMixin.js#L34
The API-Docs type {any} should be precised. Also applies for other Mixin classes.

### DIFF
--- a/app/form/LayersMixin.js
+++ b/app/form/LayersMixin.js
@@ -51,7 +51,7 @@ Ext.define('CpsiMapview.form.LayersMixin', {
     /**
      * Hides the layers associated with the form when the
      * form is hidden
-     * */
+     */
     onHide: function () {
         if (this.syncLayerWindowVisibility) {
             this.toggleLayerVisibility(false);
@@ -61,7 +61,7 @@ Ext.define('CpsiMapview.form.LayersMixin', {
     /**
      * Shows the layers associated with the form when the
      * form is hidden
-     * */
+     */
     onWindowShow: function () {
         if (this.syncLayerWindowVisibility) {
             this.toggleLayerVisibility(true);
--- a/app/form/LayersMixin.js
+++ b/app/form/LayersMixin.js
@@ -24,6 +24,11 @@ Ext.define('CpsiMapview.form.LayersMixin', {
         }
     },
 
+    /**
+     * Changes the visibility of any of the layers associated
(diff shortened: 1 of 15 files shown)


app/form/LayersMixin.js
```diff
@@ -43,12 +48,20 @@ Ext.define('CpsiMapview.form.LayersMixin', {
         }
     },
 
+    /**
+     * Hides the layers associated with the form when the
+     * form is hidden
+     * */
```
chrismayer, line 54:
```suggestion
     */
```

app/form/LayersMixin.js
```diff
@@ -43,12 +48,20 @@ Ext.define('CpsiMapview.form.LayersMixin', {
...
     onHide: function () {
         if (this.syncLayerWindowVisibility) {
             this.toggleLayerVisibility(false);
         }
     },
 
+    /**
+     * Shows the layers associated with the form when the
+     * form is hidden
+     * */
```
chrismayer, line 64:
```suggestion
     */
```

chrismayer (approved):
LGTM, just two minor formatting things.

geographika:
Thanks for the review! Merging now.

PR #505, opened 2021-09-28
Fix feature cleanup if no group prop is set

annarieger:
Properly cleanup result layer for current active group also for features which don't have property `group` set.

Needed for update of road segment features retrieved by `pmspy/edgesolver` endpoint.

Please review @geographika


geographika:
Looks good to me - merging.

PR #506, opened 2021-09-28
Introduce LengthField

annarieger:
Introduce `lengthField` component which enables simple and quick formatting of measurement fields.

By default, input value `100` becomes `100.00 m` after format is applied.

Please review @geographika


chrismayer (approved):

geographika:
Thanks @annarieger !

PR #571, opened 2022-07-08
Allow connection to protected services

geographika:
This update allows the demo application to connect to a set of development backend services. These include:

- authentication (to return a token that can then be used to access further services)
- backend Python tools to create routes

When working locally with the Sencha development server the following command should be used to run the demo app:

`sencha app watch  --j2ee`

This flag allows the `WEB-INF/web.xml` file to be used which proxies service calls to the backend development server. 
It does not seem possible to proxy back to the development server directly using https, so an additional local proxy is added, to make the following URLs available at http://localhost/WebServices and http://localhost/pmspy. 
These could be added to https://github.com/compassinformatics/cpsi-mapview/blob/master/nginx.conf and then will be available via Docker. 

On the live demo at https://compassinformatics.github.io/cpsi-mapview/#map=11/-862962.16/6899723.27/0 URLs are rewritten in the custom `Application.js` file to point to the development server directly.

### DIFF
--- a/WEB-INF/web.xml
+++ b/WEB-INF/web.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+See https://docs.sencha.com/cmd/6.5.0/guides/advanced_cmd/cmd_reference.html#advanced_cmd-_-cmd_reference_-_sencha_app_watch
+-->
+<web-app
+  xmlns="http://java.sun.com/xml/ns/javaee"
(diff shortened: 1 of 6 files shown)


geographika:
On the demo site, setting the direct URL only partially works. Calls to any protected services require the token to be passed via a cookie. As the demo site is on a separate URL from the backend services these are not allowed due to CORs. 
I'm unsure if there is anyway around this, as I presume reverse proxies are not possible on GitHub pages (maybe a custom domain could work with a proxy setup on that).

PR #576, opened 2022-07-25
Layertree: only show layers that user is allowed to see

### LINKED ISSUE #562: Hide layer tree layers dependent on user role
Currently, all layers in the layer tree are visible to all users. 

The cpsi-mapview project already contains a [CpsiMapview.util.RoleManager](https://github.com/compassinformatics/cpsi-mapview/blob/master/app/util/RoleManager.js) which reads a user's role(s) from a cookie populated by a backend authentication service.

This could be used to hide/show a layer tree node. Roles could be configured in the [tree.json](https://github.com/compassinformatics/cpsi-mapview/blob/master/resources/data/layers/tree.json) (alternatively [default.json](https://github.com/compassinformatics/cpsi-mapview/blob/master/resources/data/layers/default.json), but tree.json seems the better fit. 

Ideally, some unit tests would be added to mock this. 

Note - this update is simply to hide the layers in the UI to simplify the tree for some users, rather than to restrict access to these layers. 

image

JakobMiksch:
- fixes #562 
- works on login and logout
- also works for groups
- also tested on switch layers
- for testing:
  - set the cookie `roles` to something like `EDITOR_ROLE,DUMMY_ROLE` 
  - add the property `requiredRoles` to layer items in `tree.json`, like below :

```json
[...]
{
            "id": "LA_SITES",
            "text": "Local Authority Sites",
            "leaf": true,
            "requiredRoles": ["READER_ROLE"]
          },
          {
            "id": "RUINS_WFS",
            "leaf": true,
            "text": "Ruins",
            "qtip": "Right-click for a layer grid",
            "requiredRoles": ["ADMIN_ROLE"]
          },
          {
            "id": "my-group",
            "title": "My Group",
            "expanded": true,
            "checked": false,
            "requiredRoles": ["EDITOR_ROLE"],
            "children": [
              {
                "id": "EDGES_WFS",
                "leaf": true,
                "text": "Edges",
                "qtip": "Used in the snapping tool example"
              },
              {
                "id": "NODES_WFS",
                "leaf": true,
                "text": "Nodes",
                "qtip": "Used in the snapping tool example",
                "requiredRoles": ["ADMIN_ROLE"]
              }
            ]
          },
          {
            "id": "BOREHOLE_WMS",
            "leaf": true,
            "text": "Borehole WMS",
            "requiredRoles": ["EDITOR_ROLE"]
          },
          {
            "id": "BOREHOLE_WFS",
            "leaf": true,
            "text": "Borehole WFS (Time)",
            "qtip": "Filter records with the timeslider control",
            "requiredRoles": ["EDITOR_ROLE"]
          },
[...]
```

### DIFF
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -111,39 +111,6 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
     },
 
     /**
-     * Removes layers which the user is not allowed to see.
-     *
-     * @param {Array} children An array of layer children
-     * @return {Array} The modified children array
-     */
-    updateTreeChildrenByRole: function(children){
-        var me = this;
-        var result = [];
-        Ext.each(children, function(child){
-            var showNode = true;
-            if (child.leaf) {
-                var requiredRoles = child.requiredRoles;
-                // check if user is allowed to see layer
-                if (Ext.isArray(requiredRoles) && requiredRoles.length){
-                    showNode = false;
-                    Ext.each(requiredRoles, function(role){
-                        var userHasRole = CpsiMapview.util.RoleManager.checkRole(role);
-                        if (userHasRole){
-                            showNode = true;
-                        }
-                    });
-                }
-            } else {
-                child.children = me.updateTreeChildrenByRole(child.children);
-            }
-            if (showNode) {
-                result.push(child);
-            }
-        });
-        return result;
-    },
-
-    /**
      * This method will return an instance of the GeoExt class
      * `GeoExt.data.store.LayersTree` based on the connected OL #map. The layers
      * of the `ol.Map` are restructured and divided into groups based on the
@@ -161,7 +128,7 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
 
             // remove layers for which the user does not have the role to see it
             var modifiedTree = Ext.clone(treeJson);
-            modifiedTree.treeConfig.children = me.updateTreeChildrenByRole(modifiedTree.treeConfig.children);
+            modifiedTree.treeConfig.children = CpsiMapview.util.RoleManager.updateTreeChildrenByRole(modifiedTree.treeConfig.children);
 
             // save defaults for tree nodes from config
             me.treeConfDefaults = modifiedTree.defaults || {};
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -118,7 +118,7 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
     },
 
     /**
-     * This method will assigns an instance of the GeoExt class
+     * This method assigns an instance of the GeoExt class
      * `GeoExt.data.store.LayersTree` based on the connected OL #map to the view. The layers
      * of the `ol.Map` are restructured and divided into groups based on the
      * JSON tree structure loaded in #loadTreeStructure. This assures that
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -223,7 +223,15 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
             if (requiredRoles && Ext.isArray(requiredRoles) && requiredRoles.length){
                 var showNode = CpsiMapview.util.RoleManager.hasAtLeastOneRequiredRole(requiredRoles);
                 // needed to ensure layer shown/hidden on both tree and map
-                record.getOlLayer().setVisible(showNode);
+                // only show layer if it shall be visible initially
+                var olLayer = record.getOlLayer();
+                if (olLayer && olLayer.get('_origLayerConf') && olLayer.get('_origLayerConf').openLayers) {
+                    if (olLayer.get('_origLayerConf').openLayers.visibility) {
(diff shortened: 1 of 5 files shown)


jansule (commented):
Very nice @JakobMiksch. Just a few questions:

1. What should happen to a folder, if it does not contain any node a user can see? Should the folder be removed, should it still be visible?
2. Do we have a use case where we have to restore/update the layer tree due to e.g. role changes? If so, we might want to rather filter on the store, instead of actually modifying the original tree.

geographika:
Thanks @JakobMiksch !

In relation to the questions from @jansule:

1. Yes it would be good if a folder is empty to hide it from the user. A simpler (and possibly cleaner) solution would to be to allow the `requiredRoles` property to be set on folders directly. 
2. Ideally the function would be rerun if a user logged out and then back in. There is a global event for login and logout that can be listened to: https://github.com/compassinformatics/cpsi-mapview/blob/07e152c4e758f31c554f78125bb92d0eaa5adc58/app/controller/button/LoginButtonController.js#L85

I'm testing the branch on a client application. One thing I noticed is that in `updateTreeChildrenByRole` the following check returns false for all layer nodes as at this stage they are simple JS objects with no `leaf` property (this is only set on the folders in tree.json):

```js
//            if (child.leaf) {
// had to replace with 
            if (child.leaf !== false) {
```

geographika:
Just on the `leaf: true` - it looks like we stopped adding this in larger tree.json files, unlike in the demo application.

JakobMiksch:
Thanks @geographika @jansule for the review.
- I changed the approach from filtering the JSON to filtering the store. 
- works on login and logout
- also works for groups
- also tested on switch layers
- so far I tested the login/logout with a mock. In `app/controller/form/Login.js`, line 146 I added this:

```js
        Ext.util.Cookies.set('roles', 'EDITOR_ROLE,READER_ROLE');
        Ext.GlobalEvents.fireEvent('login', {});
        view.close();
        return;
```

JakobMiksch:
> Just on the `leaf: true` - it looks like we stopped adding this in larger tree.json files, unlike in the demo application.

I did no use the `leaf` in the new approach anymore.

geographika:
@JakobMiksch - thanks for this. Looking good when testing in a larger client application. 
One issue is that when a layer is hidden for a user, and then I log out and log in with a user who can see the layer, the layer appears in the tree but is automatically switched on (when it is not configured to be visible in `default.json`. 
I've sent details with an extra login so you may be able to create locally.

app/controller/LayerTreeController.js
```diff
@@ -202,6 +207,29 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
...
+        var store = me.getView().getStore();
+        if (!store){
+            return;
+        }
+        store.filterBy(function(record){
+            var requiredRoles = record.get('requiredRoles');
+            if (requiredRoles && Ext.isArray(requiredRoles) && requiredRoles.length){
+                var showNode = CpsiMapview.util.RoleManager.hasAtLeastOneRequiredRole(requiredRoles);
+                // needed to ensure layer shown/hidden on both tree and map
+                record.getOlLayer().setVisible(showNode);
```
geographika, line 226: Not sure if this should be the case - the node should become visible, but only turned on if set in default.json with the:

      "openLayers": {
        "visibility": true
      }

JakobMiksch:
> @JakobMiksch - thanks for this. Looking good when testing in a larger client application. One issue is that when a layer is hidden for a user, and then I log out and log in with a user who can see the layer, the layer appears in the tree but is automatically switched on (when it is not configured to be visible in `default.json`. I've sent details with an extra login so you may be able to create locally.

Good Catch! [612e488](https://github.com/compassinformatics/cpsi-mapview/pull/576/commits/612e488c4a2b0018c588ab76a433f6929732eb24) should fix it

geographika (approved):

geographika:
@JakobMiksch - I'm happy to merge this in if you are.

JakobMiksch:
@geographika I was waiting for a review by @jansule I will ping him again

app/controller/LayerTreeController.js
```diff
@@ -202,6 +207,37 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
         });
     },
 
+
+    /**
+     * Ensures tree and map only contains layers for which
```
jansule, line 212:
```suggestion
     * Ensures tree and map only contain layers for which
```

app/util/RoleManager.js
```diff
@@ -23,6 +23,22 @@ Ext.define('CpsiMapview.util.RoleManager', {
...
-    }
+    },
 
-});
\ No newline at end of file
+    /**
+     * Checks if user has at least one of the required roles.
+     *
+     * @param {String[]} requiredRoles The required roles.
+     * @return {Boolean} If user has at least one of the required roles.
```
jansule, line 32:
```suggestion
     * @return {Boolean} True, if user has at least one of the required roles. False, otherwise.
```

jansule (approved):
Very nice @JakobMiksch Just fixed some typos

JakobMiksch:
Thanks @jansule for the review, I applied the remarks.
@geographika We can merge from my side.

Issue #569, opened 2022-06-28
Allow Line geometry offsets to be created


geographika:
Add functionality to allow a new geometry to be displayed as an offset of an original line geometry.

Use case:

- A user will select a stretch of road. This will be returned by the Python routing tools as a new OL line feature
- A user will be able to create a new line as an offset of the original (to mark-up cycle tracks)

One possible option is to have number selectors on the UI that shift the new line geometry dynamically to the left or right of the line (or are 4 boxes needed to offset in 4 directions)?

![image](https://user-images.githubusercontent.com/490840/176154993-5a4efeb2-b4a8-4115-a038-e81f732d95fe.png)

![image](https://user-images.githubusercontent.com/490840/176155450-4084d3cc-7925-44a5-94ae-cd1113470bf9.png)

Review existing approaches in desktop GIS for possible UI implementations:

https://desktop.arcgis.com/en/arcmap/latest/manage-data/creating-new-features/creating-a-segment-parallel-to-another-segment.htm

https://gis.stackexchange.com/questions/229914/creating-parallel-lines-at-particular-distance-from-lines-which-are-part-of-a-sh

https://github.com/stdonn/ParallelLineConstruction

jansule:
@geographika Are there any limitations for adding additional javascript dependencies to the project? I was thinking about adding [turf.js lineOffset](https://turfjs.org/docs/#lineOffset) (MIT license), for not having to (re-)implement the offsetting logic by myself.

geographika:
@jansule - it looks like just this function can be added using `npm install @turf/line-offset` - is this correct?
The bulk of the code is https://github.com/Turfjs/turf/blob/master/packages/turf-line-offset/index.js
I think if this works well and is fully-tested it makes sense to use this package (and add to package.json?).

jansule:
> it looks like just this function can be added using npm install @turf/line-offset - is this correct?

Yes, this is correct.

> I think if this works well and is fully-tested it makes sense to use this package (and add to package.json?).

Alright, thanks. I will add this to package.json and will update app.json accordingly.

geographika:
Thanks @jansule - closing.

PR #577, opened 2022-07-27
Parallel line

jansule:
This adds the ParallelLine toolbar.

This toolbar expects a feature with a line geometry and provides an input field for specifying an offset and a trigger button to create a new feature that is parallel to the provided feature. The toolbar only fires an event `parallelLineCreated` that contains the newly created feature. Any handling with regard to adding the feature to a layer must be done in the wrapping component. `ParallelLineWindow` provides an example implementation for that.

Once a parallel line was created for the given feature, any changes in the offset will automatically fire the `parallelLineCreated` event again.

Example:

compass-parallel-line

Please note that depending on the offset and the feature's geometry, the geometry of the parallel feature might differ from its original geometry.

Example:

compass-parallel-line-high-offset

This solves https://github.com/compassinformatics/cpsi-mapview/issues/569

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/app/controller/window/ParallelLineWindow.js
+++ b/app/controller/window/ParallelLineWindow.js
@@ -1,0 +1,59 @@
+/**
+ * This class is the controller for the ParallelLine.
+ */
+Ext.define('CpsiMapview.controller.window.ParallelLine', {
+    extend: 'Ext.app.ViewController',
+
+    alias: 'controller.cmv_parallel_line_window',
+
+    requires: [
+        'BasiGX.view.component.Map'
+    ],
+
+    selectInteraction: null,
+
+    layer: null,
+
+    onSelectFeatureToggle: function (btn, toggled) {
+        var me = this;
+        var view = me.getView();
+        var vm = view.getViewModel();
+        var map = BasiGX.view.component.Map.guess().getMap();
+        if (toggled) {
+            var interactionConfig = view.interactionConfig ? view.interactionConfig : {};
+            this.selectInteraction = new ol.interaction.Select(interactionConfig);
+            this.selectInteraction.on('select', function (evt) {
+                var selectedFeature = evt.selected[0];
+                vm.set('selectedFeature', selectedFeature);
+            });
+            map.addInteraction(this.selectInteraction);
+
+        } else {
+            if (!this.selectInteraction) {
+                return;
+            }
+            map.removeInteraction(this.selectInteraction);
+            this.selectInteraction = null;
+        }
+    },
+
+    onParallelLineCreated: function (newLineFeature) {
+        if (this.layer === null) {
+            this.layer = new ol.layer.Vector();
+            var map = BasiGX.view.component.Map.guess().getMap();
+            if (map) {
+                map.addLayer(this.layer);
+            }
+        }
+        var source = new ol.source.Vector({
+            features: [newLineFeature]
+        });
+        this.layer.setSource(source);
+    },
+
+    onBeforeDestroy: function () {
+        // remove select interaction
+        this.onSelectFeatureToggle(null, false);
+    }
+
+});
--- a/app/view/toolbar/ParallelLineToolbar.js
+++ b/app/view/toolbar/ParallelLineToolbar.js
@@ -1,0 +1,165 @@
+/**
+ * This class is the toolbar for creating parallel lines.
+ * @class CpsiMapview.view.toolbar.ParallelLine
+ */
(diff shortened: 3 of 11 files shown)


.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
```
JakobMiksch, line 1: is this file relevant for the PR?

app/view/toolbar/ParallelLineToolbar.js
```diff
@@ -0,0 +1,165 @@
...
+            return;
+        }
+
+        var offset = vm.get('offset');
+        if (offset === null) {
+            return;
+        }
+
+        var offsetUnit = this.getOffsetUnit();
+        var map = BasiGX.view.component.Map.guess().getMap();
```
JakobMiksch, line 151: just wondering if we could move the part, which is not dependent on the component, to a util for "turf" functionality. It would be easier for testing and might be reused somewhere else.

app/controller/window/ParallelLineWindow.js
```diff
@@ -0,0 +1,59 @@
...
+    onSelectFeatureToggle: function (btn, toggled) {
+        var me = this;
+        var view = me.getView();
+        var vm = view.getViewModel();
+        var map = BasiGX.view.component.Map.guess().getMap();
+        if (toggled) {
+            var interactionConfig = view.interactionConfig ? view.interactionConfig : {};
+            this.selectInteraction = new ol.interaction.Select(interactionConfig);
+            this.selectInteraction.on('select', function (evt) {
+                var selectedFeature = evt.selected[0];
```
JakobMiksch, line 26: is it possible that the user accidentally selects two features? this could maybe be handled

JakobMiksch (commented):
@jansule Code looks fine, a few questions/notes inside.

.editorconfig
jansule: It is relevant in the sense that it automatically sets the right/consistent editor configurations. But this could be moved in a separate PR of course

app/controller/window/ParallelLineWindow.js
jansule: As described in the PR description, `ParallelLineWindow` is just an example implementation and not intended for production use. The actual component is `PrallelLine` toolbar.

JakobMiksch: alright

app/view/toolbar/ParallelLineToolbar.js
jansule: Sure, I can refactor this and create a util for that

jansule: done in https://github.com/compassinformatics/cpsi-mapview/pull/577/commits/75cf1cc77e4a8316bb0557e9e67644d7ce15d8eb

geographika (approved):

geographika:
@jansule - looks great, thanks!

A couple of questions:

1. It looks like the layer style etc. can be set in the calling code, as in the window example - is this correct?
2. Are the offset lines always the same length as the original?

There might be some refinement to the tool once we get it setup in a live project, but this is a great start. 
One possible addition is to check that the offset line remains a (valid) LineString, to catch when it breaks in two, as in your example above. Then a tool could check for this validity prior to saving.

geographika:
Thanks for the great explanation @jansule. Any thoughts on updating the `console` calls? Apart from that, I'm happy to merge and close the issue if you are.

jansule:
> Any thoughts on updating the `console` calls?

I will replace the console.logs (tomorrow :grinning:). AFAIK Ext.log is not part of a production build, so these logs would disappear in production environments. Is this intended behavior from your side?

geographika:
I think it just defaults to `console` when available - "Logs a message. If a console is present it will be used. " - https://docs.sencha.com/extjs/6.7.0/classic/Ext.html#method-log

I know the `Ext.Assert` don't make it to production builds and have to be wrapped, but I thought `Ext.log` did (although we typically only review console messages when developing locally). 

```js
        //<debug>
        Ext.Assert.truthy(finalFilters.length <= 2);
        //</debug>
```

JakobMiksch (approved):
I ran it again locally and it works fine for me.

app/view/toolbar/ParallelLineToolbar.js
```diff
@@ -0,0 +1,157 @@
...
+        }
+        var geometry = feature.getGeometry();
+        if (!geometry) {
+            console.warn('Feature has no geometry.');
+            return;
+        }
+
+        var geomType = geometry.getType();
+        if (geomType !== 'LineString' && geomType !== 'MultiLineString') {
+            console.warn('Unsupported feature geometry. Geometry must be of type "LineString" or "MultiLineString".');
```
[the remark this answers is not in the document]
jansule: Done in https://github.com/compassinformatics/cpsi-mapview/pull/577/commits/85c2183296cdf7c2207b179a99c1fd130e9d4c20

jansule:
Thanks for the reviews. Replaced the logs. Will merge now.

Issue #564, opened 2022-06-14
Drawing Tool Improvements


geographika:
The drawing tool #556 is functional, but its usability could be improved with some/all of the following features:

- adding a symbol at the start and end of the line

![image](https://user-images.githubusercontent.com/490840/173599729-ea322e85-9afd-45bc-a1c0-3a03a4532f9a.png)

- giving a user feedback if the start or end of the line has snapped to a feature - possibly by applying a fill colour to the symbols at the start and end of the line? Maybe "flashing" the feature the line snaps to would also aid a user in ensuring they are snapping to the correct feature

- (just a suggestion at this stage) allowing the line to be selected and showing the vertices that make up the line

A demo of the existing drawing tool can be given to show its current usage.

geographika:
Thanks @JakobMiksch!

PR #578, opened 2022-07-27
Drawing Tool Improvements

JakobMiksch:
- references #564 
- snapped node or edges are highlighted
- provides various configurable styles for editing
- todo: 
  - [x] handle polygons

## snapping

Peek 2022-07-27 18-55

## highlight vertices of snapped line

Peek 2022-07-28 10-25

### DIFF
--- a/app/controller/button/DrawingButtonController.js
+++ b/app/controller/button/DrawingButtonController.js
@@ -169,19 +169,20 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
             // remember if we have hit a referenced layer
             var nodeLayerHit = false;
             var edgeLayerHit = false;
+            var polygonLayerHit = false;
             var selfHit = false;
 
             var snappedEdgeFeature;
             me.map.forEachFeatureAtPixel(pixel, function (foundFeature, layer) {
+                snappedEdgeFeature = foundFeature;
                 if (layer) {
                     var key = layer.get('layerKey');
                     if (key === view.getNodeLayerKey()) {
                         nodeLayerHit = true;
                     } else if (key === view.getEdgeLayerKey()) {
                         edgeLayerHit = true;
-                        snappedEdgeFeature = foundFeature;
                     } else if (key === view.getPolygonLayerKey()) {
-                        // TODO: handle polygon hit
+                        polygonLayerHit = true;
                     } else if (me.drawLayer === layer) {
                         // snapping to self drawn feature
                         selfHit = true;
@@ -213,6 +214,8 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
                 } else {
                     return view.getSnappedEdgeStyle();
                 }
+            } else if (polygonLayerHit) {
+                return view.getSnappedEdgeStyle();
             } else if (selfHit) {
                 return view.getModifySnapPointStyle();
             } else {
--- a/app/controller/button/DrawingButtonController.js
+++ b/app/controller/button/DrawingButtonController.js
@@ -45,7 +45,7 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
     snapInteraction: null,
 
     /**
-     * If user has started to edit a line i.d. the first point of a line is already set
+     * If user has started to edit a line, this means the first point of a line is already set
      */
     editingIsActive: false,
 
--- a/app/controller/button/DrawingButtonController.js
+++ b/app/controller/button/DrawingButtonController.js
@@ -179,24 +179,42 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
             var nodeLayerHit = false;
             var edgeLayerHit = false;
             var selfHit = false;
-            me.map.forEachFeatureAtPixel(pixel, function (feature, layer) {
+            var snappedEdgeFeature;
+            me.map.forEachFeatureAtPixel(pixel, function (foundFeature, layer) {
                 if (layer) {
                     var key = layer.get('layerKey');
                     if (key === view.getNodeLayerKey()) {
                         nodeLayerHit = true;
                     } else if (key === view.getEdgeLayerKey()) {
                         edgeLayerHit = true;
+                        snappedEdgeFeature = foundFeature;
+                    } else if (key === view.getPolygonLayerKey()) {
+                        // TODO: handle polygon hit
                     } else if (me.drawLayer === layer) {
                         // snapping to self drawn feature
                         selfHit = true;
                     }
-                    // TODO: handle polygons
                 }
             });
             if (nodeLayerHit) {
                 return view.getSnappedNodeStyle();
             } else if (edgeLayerHit) {
-                return view.getSnappedEdgeStyle();
+                if (!snappedEdgeFeature) {
+                    return;
+                }
+
+                // Prepare style for vertices of snapped edge
+                var geom = snappedEdgeFeature.getGeometry();
+                var coords = geom.getCoordinates();
+                var verticesMultiPoint = new ol.geom.MultiPoint(coords);
+                var vertexStyle = view.getSnappedEdgeVertexStyle();
+                vertexStyle.setGeometry(verticesMultiPoint);
+
+                // combine style for snapped point and vertices of snapped edge
+                return [
+                    vertexStyle,
+                    view.getSnappedEdgeStyle()
+                ];
             } else if (selfHit) {
                 return view.getModifySnapPointStyle();
             } else {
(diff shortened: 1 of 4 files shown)


JakobMiksch:
@geographika this PR so far provides a default style. I guess there is a need to manually adjust the Style via the configs in `CpsiMapview.view.tool.DrawingButton`

possible options would be:
- start/end point
  - color
  - shape
  - size
- line 
  - color
  - width

shall we expose these properties as config? Which one would be relevant here?

geographika:
@JakobMiksch can we do similar to https://github.com/compassinformatics/cpsi-mapview/blob/f12f00982031f167c74c9316cc10c01ff73e7ab2/app/view/button/DigitizeButton.js#L54 and allow a full style object to be passed in as a config?

JakobMiksch:
@geographika I updated the PR. I added the configs for the styles in the way you suggested. 
Also I created a user feedback if the edges or node layers are snapped.
Could you tell me if this is the editing behavior you are expecting?

geographika:
@JakobMiksch - these updates are great - thanks!
The symbol display when hovering over a snapped feature works very well. 

If I wanted to keep the same start and end symbology once the line is created, would I need to apply the style to the Edges vector layer?

Style when drawing:

image

Style when completed:

image

JakobMiksch:
@geographika I added highlighting for the vertices of the snapped edge. See the new gif in this PR's description. The code is pushed. What is your impression?
(I saw your comment above already, and will check it now)

geographika:
Very nice! Showing the vertices will be useful in a number of scenarios. Could we make it configurable via the view on whether vertices should be shown? The tool will be reused in a few places, so being able to turn off vertices would be useful.

app/controller/button/DrawingButtonController.js
```diff
@@ -44,13 +44,19 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
      */
     snapInteraction: null,
 
+    /**
+     * If user has started to edit a line i.d. the first point of a line is already set
```
geographika, line 48: Not sure what i.d. here means?

app/controller/button/DrawingButtonController.js
```diff
@@ -97,19 +103,126 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
...
             type: 'LineString',
-            features: layer.getSource().getFeaturesCollection(),
+            features: collection,
             condition: drawCondition,
+            style: me.getDrawStyleFunction(),
             snapTolerance: 2 // Pixel distance for snapping to the drawing finish (default 12)
         };
 
         me.drawInteraction = new ol.interaction.Draw(drawInteractionConfig);
+        me.drawInteraction.on('drawstart',me.handleDrawStart);
```
geographika, line 119: Needs a space after comma to pass linting?

geographika (commented):
Only very minor comments

app/controller/button/DrawingButtonController.js
JakobMiksch: fixed with [abc05db](https://github.com/compassinformatics/cpsi-mapview/pull/578/commits/abc05dba3e376f274ce2ecd1c8be494da91234b0)

JakobMiksch:
> @JakobMiksch - these updates are great - thanks! The symbol display when hovering over a snapped feature works very well.
> 
> If I wanted to keep the same start and end symbology once the line is created, would I need to apply the style to the Edges vector layer?
> 
> Style when drawing:
> 
> image
> 
> Style when completed:
> 
> image

@geographika in https://github.com/compassinformatics/cpsi-mapview/pull/578/commits/76e6a632845578a21c3d36ba1a3c375b027b2907 I added the drawStyle as the default style of the layer. Should this be the default? And should it be possible to configure the default style of the drawn layer, once it is not in editing mode anymore?

app/controller/button/DrawingButtonController.js
JakobMiksch: I actually meant `i.e.`, but I simply replaced it with "this means". This might be better understandable.
Fixed with [8a1a771](https://github.com/compassinformatics/cpsi-mapview/pull/578/commits/8a1a7711b7d95d6f15d6ab4b2f9a073733558efd)

geographika: I tried to google the equivalent of e.g. in German to see if it was that, should have thought of i.e. first!

geographika:
> @geographika in [76e6a63](https://github.com/compassinformatics/cpsi-mapview/commit/76e6a632845578a21c3d36ba1a3c375b027b2907) I added the drawStyle as the default style of the layer. Should this be the default? And should it be possible to configure the default style of the drawn layer, once it is not in editing mode anymore?

I'm not sure if I fully understand the quests here, but I've tested the last update, and this approach is perfect. Keeping the Start and End styles of the line once created is the main requirement (so a user know which end is which). 

If we could make hiding/showing vertices configurable, I think this update is then ready to merge?

JakobMiksch:
@geographika do we need to test/implement the snapping for polygon layers as well?

geographika:
> @geographika do we need to test/implement the snapping for polygon layers as well?

Ideally, yes, but the demo currently doesn't have a polygon layer to test with. Would the approach be very different from polygons compared to lines?
Are you able to add in the code without testing? I could then test the branch in a local project with polygons. Otherwise, I can add a sample polygon layer to the demo.

JakobMiksch:
I assume, polygons would be different from lines, because we might need to check if the cursor is on the boundary or the inside of the polygon.

JakobMiksch:
In 06fb051 I added minimal support for polygons (without testing). Maybe you could try if this is already okay for you? 
Otherwise I would need to refine it.

geographika:
@JakobMiksch - I've tested the branch locally with a polygon layer and it highlights the snap correctly:

image

One minor issue is that the vertices are not visible above the polygon layer - does it need to be moved to the top of the rendering stack?

When polygon layer is on:

image

When polygon layer is off (the white points are now visible):

image

JakobMiksch:
@geographika It indeed looks like the highlighted vertices are rendered too low.
Could you try adding this line: `snappedEdgeVertexStyle.setZIndex(Infinity);` in line 208 of `app/controller/button/DrawingButtonController.js`?
This might be an easy fix

app/view/button/DrawingButton.js
```diff
@@ -60,7 +60,150 @@ Ext.define('CpsiMapview.view.tool.DrawingButton', {
...
+        /**
+         * If the vertices of the snapped edge shall be shown.
+         */
+        showVerticesOfSnappedEdge: true,
+
+        /**
+         * Style of the editing cursor, before first point is set.
+         */
+        drawBeforeEditingPoint: new ol.style.Style({
+            image: new ol.style.Circle({
```
jansule, line 74: Minor: Styles could be refactored into a dedicated class, in order to ensure better readability of the `DrawingButton` class

jansule (approved):
Nice work @JakobMiksch. I'm wondering if it makes sense to provide this functionality in BasiGX, since it might be useful for others as well.

app/view/button/DrawingButton.js
JakobMiksch: Good idea. I did it with https://github.com/compassinformatics/cpsi-mapview/pull/578/commits/78769d6a4f510b9ffe8b29cbeef59e3795a98f8d

JakobMiksch:
Thanks @jansule thanks for the review. I moved the styles to a util.
Moving the code to BasiGX would be fine for me. But first I would commit the code to this repo.

JakobMiksch:
@geographika I moved to styles to a util and made sure they are always created as new instance. My suspicion is that the bug with the vertices rendering might have to do with that. Maybe you can try again now. Otherwise I think I would be best if I had a polygon layer to test for my self.

geographika:
> @geographika I moved to styles to a util and made sure they are always created as new instance. My suspicion is that the bug with the vertices rendering might have to do with that. Maybe you can try again now. Otherwise I think I would be best if I had a polygon layer to test for my self.

@JakobMiksch - I've added a polygon layer to the example in https://github.com/compassinformatics/cpsi-mapview/pull/579
It is a minor issue, but if you hover over a line within a polygon the vertices don't show:
CpsiMapview

JakobMiksch:
Thanks  @geographika for providing the polygon layer. It helped me to find the problem. In the situation of the described bug, not the line feature was returned anymore but the polygon feature. This is fixed now.
I also added a new style ( a black star) indicating when the cursor is snapping or above the referenced polygon layer.

geographika:
Tested locally and it works very well with all layer types. If you are happy @JakobMiksch I can go ahead and merge this.

JakobMiksch:
@geographika Great to hear that it works! This PR can be merged from my side.

PR #580, opened 2022-07-29
Add filter to layer tree

### LINKED ISSUE #484: Allow Searching of Layer Tree
Add similar functionality as shown in the Sencha Kitchen Sink app: https://examples.sencha.com/extjs/7.4.0/examples/kitchensink/?classic#direct-tree

Sencha _ Ext JS Examples

An additional feature would be to add a checkbox titled "Show Visible", which would filter the tree based on if the layer's visibility is set to true.

JakobMiksch:
- fixes #484
- adds textfield for filtering by layer name
- adds checkbox to filter by visible layers
- adds a config if baselayers should be filtered or not (defaults to `true`)

Peek 2022-07-29 14-28

### DIFF
--- a/app/form/LayerTreeFilter.js
+++ b/app/form/LayerTreeFilter.js
@@ -37,25 +37,24 @@ Ext.define('CpsiMapview.controller.form.LayerTreeFilter', {
         var filter = Ext.util.Filter({
             id: view.FILTER_ID,
             filterFn: function(node) {
-                if (vm.get('searchText')) {
-                    var isMatch = LayerTreeFilterUtil.isSearchTextInLayerName(node, vm.get('searchText'), filterBaseLayers);
-                    if (!isMatch) {
-                        return false;
+                var isLeaf = !node.hasChildNodes();
+                if (isLeaf) {
+                    if (vm.get('hideInvisibleLayers')) {
+                        var isVisible = LayerTreeFilterUtil.isLayerVisible(node, filterBaseLayers);
+                        if (!isVisible) {
+                            return false;
+                        }
                     }
-                }
-                if (vm.get('hideInvisibleLayers')) {
-                    var isVisible = LayerTreeFilterUtil.isLayerVisible(node, filterBaseLayers);
-                    if (!isVisible) {
-                        return false;
+                    if (vm.get('searchText')) {
+                        var isMatch = LayerTreeFilterUtil.isSearchTextInLayerName(node, vm.get('searchText'), filterBaseLayers);
+                        if (!isMatch) {
+                            return false;
+                        }
                     }
+                    return true;
+                } else {
+                    return false;
                 }
-                if (view.hideGroupsWithNoVisibleLayer) {
-                    var keepGroup = LayerTreeFilterUtil.groupHasVisibleLayer(node);
-                    if (!keepGroup) {
-                        return false;
-                    }
-                }
-                return true;
             }
         });
         store.addFilter(filter);
@@ -76,9 +75,6 @@ Ext.define('CpsiMapview.controller.form.LayerTreeFilter', {
             return;
         }
         vm.set('searchText', searchText);
-        // Making sure we work on the latest state,
-        // when updating the filter.
-        vm.notify();
         ctrl.updateFilter();
     },
 
@@ -97,9 +93,6 @@ Ext.define('CpsiMapview.controller.form.LayerTreeFilter', {
             return;
         }
         vm.set('hideInvisibleLayers', checked);
-        // Making sure we work on the latest state,
-        // when updating the filter.
-        vm.notify();
         ctrl.updateFilter();
     }
 });
--- a/app/form/LayerTreeFilter.js
+++ b/app/form/LayerTreeFilter.js
@@ -1,0 +1,103 @@
+/**
+ * This is the controller for the {@link CpsiMapview.view.form.LayerTreeFilter} component
+ *
+ * @class CpsiMapview.controller.form.LayerTreeFilter
+ */
+Ext.define('CpsiMapview.controller.form.LayerTreeFilter', {
+    extend: 'Ext.app.ViewController',
+
+    alias: 'controller.cmv_layertreefilter',
+
+    /**
+     * Filters the tree by matching the layer's names with the input text.
+     *
+     * @param {Ext.form.field.Text} textField The textfield
+     * @param {String} newVal The new value
+     */
+    onSearch: function (textField, newVal) {
+        var me = this;
+        var view = me.getView();
+
+        // show/hide clear trigger button
+        if (textField.getValue()) {
+            textField.setHideTrigger(false);
+        } else {
+            textField.setHideTrigger(true);
+        }
+
+        var tree = Ext.ComponentQuery.query('cmv_layertree')[0];
+        if (!tree){
+            return;
+        }
+        var store = tree.getStore();
+        if(!store){
+            return;
+        }
+
+        // remove previous filter if exists
+        store.removeFilter(view.TEXT_FILTER_ID);
+
+        var textFilter = new Ext.util.Filter({
+            id: view.TEXT_FILTER_ID,
+            filterFn: function (node) {
+                // always show group layers
+                if (node.hasChildNodes()) {
+                    return true;
+                }
+                // check if baselayers should be filtered
+                if (!view.getDoFilterBaseLayers() && node.getOlLayerProp('isBaseLayer')){
+                    return true;
+                }
+                // prevent case sensitive matches
+                var name = node.getOlLayerProp('name');
+                if (!name) {
+                    return;
+                }
+                var text = name.toLowerCase();
+                var compare = newVal.toLowerCase();
+                return text.includes(compare);
+            }
+        });
+        store.addFilter(textFilter);
+    },
+
+    /**
+     * Filters the tree by checking the visibility of the layers.
+     *
+     * @param {Ext.form.field.Field} checkBox The checkbox
+     * @param {Boolean} showOnlyVisible If only visible layers should be shown in the tree
+     */
+    filterVisibleLayers: function (checkBox, showOnlyVisible) {
+        var me = this;
+        var view = me.getView();
+
+        var tree = Ext.ComponentQuery.query('cmv_layertree')[0];
+        if (!tree) {
+            return;
+        }
+        var store = tree.getStore();
+        if (!store) {
+            return;
+        }
+        if (showOnlyVisible) {
+            var visibleLayerFilter = new Ext.util.Filter({
+                id: view.VISIBLE_LAYER_FILTER_ID,
+                filterFn: function (node) {
+                    if (node.hasChildNodes()) {
+                        return true;
+                    }
+                    // check if baselayers should be filtered
+                    if (!view.getDoFilterBaseLayers() && node.getOlLayerProp('isBaseLayer')){
+                        return true;
+                    }
+                    var checked = node.get('checked');
+                    return checked;
+                }
+            });
+
+            store.addFilter(visibleLayerFilter);
+        } else {
+            store.removeFilter(view.VISIBLE_LAYER_FILTER_ID);
+        }
+    }
+});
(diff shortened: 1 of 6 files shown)


app/view/form/LayerTreeFilter.js
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * The ID of the filter created by the text field.
+     */
+    TEXT_FILTER_ID: 'cpsi-tree-layer-text-filter',
+
+    /**
+     * The ID of the filter created by the checkbox.
+     */
+    VISIBLE_LAYER_FILTER_ID: 'cpsi-tree-visible-layer',
```
JakobMiksch, line 32:
```suggestion
    VISIBLE_LAYER_FILTER_ID: 'cpsi-tree-visible-layer-filter',
```

app/form/LayerTreeFilter.js
```diff
@@ -0,0 +1,103 @@
...
+            filterFn: function (node) {
+                // always show group layers
+                if (node.hasChildNodes()) {
+                    return true;
+                }
+                // check if baselayers should be filtered
+                if (!view.getDoFilterBaseLayers() && node.getOlLayerProp('isBaseLayer')){
+                    return true;
+                }
+                // prevent case sensitive matches
```
jansule, line 51:
```suggestion
                // enforce case insensitive matches
```

jansule (commented):
Nice work @JakobMiksch.

Just one question: Should empty folders also be filtered out?

JakobMiksch:
Thanks for the review. 
Regarding empty directories and general usability:
@geographika what is your impression?

geographika:
Filtering works very well - thanks @JakobMiksch !

> Thanks for the review. Regarding empty directories and general usability: @geographika what is your impression?

In some dependent projects filtering out empty folders would greatly help usability - if it is possible to add this, please do. An alternative could be to expand/collapse all nodes, but I prefer the option of hiding them. 

In other testing, sometimes the node text contains the style name:

image

Is it easy to filter by the text on the node rather than the layer name itself?

app/form/LayerTreeFilter.js
```diff
@@ -0,0 +1,103 @@
...
+                // always show group layers
+                if (node.hasChildNodes()) {
+                    return true;
+                }
+                // check if baselayers should be filtered
+                if (!view.getDoFilterBaseLayers() && node.getOlLayerProp('isBaseLayer')){
+                    return true;
+                }
+                // enforce case insensitive matches
+                var name = node.getOlLayerProp('name');
```
geographika, line 52: As in comment - perhaps checking against the node text is more user friendly?

app/view/form/LayerTreeFilter.js
```diff
@@ -0,0 +1,59 @@
...
+    VISIBLE_LAYER_FILTER_ID: 'cpsi-tree-visible-layer-filter',
+
+    items: [{
+        xtype: 'textfield',
+        emptyText: 'Search for layers',
+        hideTrigger: true,
+        anchor: '100%',
+        maxWidth: 250,
+        triggers: {
+            clearText: {
```
geographika, line 41: Out of interest - when is this function called?

JakobMiksch: The function is called when the user clicks on the cross. The name `clearText` is just a custom name (see https://docs.sencha.com/extjs/6.2.0/classic/Ext.form.field.Text.html#cfg-triggers). It could be any other name. In fact, with 051afdd I changed its name to `clearTreeFilterText` to make it more expressive.

geographika: Thanks for the explanation. I hadn't realised `trigger` was the name ExtJS uses to refer to the button associated with the field and was looking for a custom event.

app/form/LayerTreeFilter.js
JakobMiksch: yes, that works. Done in 7e4acb0

geographika: Thanks!

JakobMiksch:
I added a filter that ensures that empty groups are hidden as well.

Can be tested with this  `tree.conf` :

```json
{
  "defaults": {
    "general": {}
  },
  "treeConfig": {
    "id": "root",
    "children": [
      {
        "id": "overlay-folder",
        "title": "Layers",
        "expanded": true,
        "checked": false,
        "children": [
          {
            "id": "LA_SITES",
            "text": "Local Authority Sites",
            "leaf": true
          },
          {
            "id": "RUINS_WFS",
            "leaf": true,
            "text": "Ruins",
            "qtip": "Right-click for a layer grid"
          },
          {
            "id": "test-dir",
            "title": "TEST",
            "expanded": true,
            "checked": false,
            "children": [
              {
                "id": "NODES_WFS",
                "leaf": true,
                "text": "Nodes",
                "qtip": "Used in the snapping tool example"
              },
              {
                "id": "EDGES_WFS",
                "leaf": true,
                "text": "Edges",
                "qtip": "Used in the snapping tool example"
              },
              {
                "id": "test-dir2",
                "title": "TEST2",
                "expanded": true,
                "checked": false,
                "children": [
                  {
                    "id": "OSM_WMS",
                    "leaf": true,
                    "text": "OSM WMS",
                    "qtip": "An OSM based WMS layer",
                    "glyph": "ea68@font-gis"
                  },
                  {
                    "id": "OSM_WMS2",
                    "leaf": true,
                    "text": "A WMS 2",
                    "glyph": "ea9c@font-gis"
                  },
                  {
                    "id": "OSM_WMS3",
                    "leaf": true,
                    "text": "A WMS 3",
                    "glyph": "ea9c@font-gis"
                  },
                  {
                    "id": "OSM_WMS4",
                    "leaf": true,
                    "text": "A WMS 4",
                    "glyph": "ea9c@font-gis"
                  }
                ]
              }
            ]
          },
          {
            "id": "BOREHOLE_WMS",
            "leaf": true,
            "text": "Borehole WMS"
          },
          {
            "id": "BOREHOLE_WFS",
            "leaf": true,
            "text": "Borehole WFS (Time)",
            "qtip": "Filter records with the timeslider control"
          },
          {
            "id": "WATERBODY_SWITCH_LAYER_FAR",
            "leaf": true,
            "text": "Waterbody Switch Layer (far)"
          },
          {
            "id": "WATERBODY_SWITCH_LAYER_CLOSE",
            "leaf": true,
            "text": "Waterbody Switch Layer (close)"
          },
          {
            "id": "TIME_WFS_PROXY",
            "leaf": true
          },
          {
            "id": "WORKS_EXPORT_WMS_PROXY",
            "leaf": true
          },
          {
            "id": "LIGHT_UNIT_SWITCH_LAYER_FAR",
            "leaf": true
          },
          {
            "id": "WATERWAYS_VTWMS",
            "leaf": true,
            "text": "Waterways VT WMS"
          },
          {
            "id": "GAS_WFS",
            "leaf": true,
            "text": "GAS WFS",
            "glyph": "ea9e@font-gis"
          },
          {
            "id": "COUNTRY_WFS",
            "leaf": true,
            "text": "Country WFS",
            "qtip": "An OSM based WFS layer",
            "glyph": "ea9e@font-gis"
          }
        ]
      },
      {
        "id": "base-folder",
        "title": "Base Layers",
        "expanded": true,
        "children": [
          {
            "id": "OSM_BACKGROUND",
            "leaf": true,
            "text": "OpenStreetMap",
            "qtip": "An OpenStreetMap based background layer",
            "cls": "cpsi-tree-node-baselayer"
          },
          {
            "id": "GREY_BACKGROUND",
            "leaf": true,
            "text": "Grey Background",
            "qtip": "This is the background layer",
            "cls": "cpsi-tree-node-baselayer"
          }
        ]
      }
    ]
  }
}
```

app/view/form/LayerTreeFilter.js
```diff
@@ -0,0 +1,109 @@
...
+                    });
+                    return keep;
+                } else {
+                    return true;
+                }
+            }
+        });
+
+        // we have to add this filter after the store of the layer tree is created.
+        // this needs two steps:
```
jansule, line 87:
```suggestion
        // this needs following steps:
```

jansule (approved):
Nice work @JakobMiksch! Just a minor non-blocking comment

JakobMiksch:
@geographika I updated the PR to also take groups into account.

app/view/form/LayerTreeFilter.js
```diff
@@ -0,0 +1,109 @@
...
+
+        // we need to set a filter that hides groups that do not have visible children
+        var emptyGroupFilter = Ext.util.Filter({
+            id: me.HIDE_EMPTY_GROUPS_FILTER_ID,
+            filterFn: function (node) {
+                if (node.hasChildNodes()) {
+                    var keep = false;
+                    node.cascade(function (cascadeNode) {
+                        // search for any descending node that is checked
+                        if (!cascadeNode.hasChildNodes() && cascadeNode.get('checked')) {
```
geographika, line 75: Just to confirm the expected behaviour here is that if layers are switched on then their parent folders will be displayed for any searches?

So for example the Road Schedule layer is turned on (in the "Road Network Layers" folder) and then a search is made which does not match any layers in this folder, the folder remains visible (and cannot be expanded):

Step 1:

image

Step 2:

image

Would it make sense to hide empty folders when searching, even if they contain a visible layer?

geographika:
@JakobMiksch - just one comment in the code review. I think it makes sense to hide folders which have no layers returned via a search, but as a layertree can have multiple hierarchies I'm not sure how feasible it is to implement. Let me know your thoughts on the comment, and if it is lots of effort to implement the logic.

app/view/form/LayerTreeFilter.js
JakobMiksch: @geographika yes true. your described behaviour is correct.

Good question, this could indeed useful.  I will check how much effort that is.

JakobMiksch:
@geographika I added a config `hideGroupsWithNoVisibleLayer`. The current behavior stays the same. Now it is easier to test the effect of this filter, by switching it of.

geographika:
> @geographika I added a config `hideGroupsWithNoVisibleLayer`. The current behavior stays the same. Now it is easier to test the effect of this filter, by switching it of.

@JakobMiksch - not sure I'd ever have this turned off, but good to see the difference! My question was more around hiding folders which have no search results, but as they have a visible layer in them they are still visible. 

In the demo version with you tree.json from above - I've turned all layers off except for "Edges", and a base layer:

image

Now if I make a search for "Ruins" the "TEST" folder is still visible (but has no layers and cannot be expanded), as is "Base Layers". Could these hidden too?

image

jansule:
@geographika I updated the code, so that now, the filter should (hopefully) work as expected.

initial state:
image

using search term (folder `TEST2` disappears):
image

using search term and hide invisible layers (folder `TEST` disappears)
image

geographika:
@jansule - had a test with the latest code and the tree.json - it is almost working as I thought it would. There still seem to be empty folders visible (in the demo it is not a huge issue, but in larger applications based on MapView there can sometimes be lots of empty folder after filtering. 

Example:

- Turn off all layers except for "A WMS4"
image
- Now search for "Nodes". I would expect the "TEST 2" folder to be hidden - regardless of the "Show visible only" setting. There are no layers with "Nodes" in the title, and the folder only seems to be displayed as it has a layer that was turned on:
image

Does the above make sense?

jansule:
This might be a timing problem with the ExtJS lifecycle. I added a function call to enforce synchronous state in the viewmodel. Could you check again, if it works now?

geographika:
> This might be a timing problem with the ExtJS lifecycle. I added a function call to enforce synchronous state in the viewmodel. Could you check again, if it works now?

Unfortunately I still see the folder even with the `notify()` additions. Is this not the same for you locally?

image

jansule:
It is the same. I was not exactly following the described workflow.

geographika:
@jansule - I just tried with your last commit - this gives the results I was expecting. If you are happy please merge. Thanks to both of you for persisting with this.

jansule:
@geographika that's great to hear. I am currently doing some code cleanup and I will try to create some meaningful unit tests, so that functionality can be ensured.

Issue #581, opened 2022-07-29
Layers with requiredRole turning off automatically


geographika:
@JakobMiksch - something that has come to light in testing #576 is that the filter function runs for all nodes in the tree and switches off layers unexpectedly. See example below, both layers have a `requiredRole` property set, but turning on the second layer turns off the first:

![Pavement Management System (5)](https://user-images.githubusercontent.com/490840/181780407-3186b601-d0d7-43ea-b711-0cfd0eb082c7.gif)

Relevant code below. It looks like visibility gets set to its definition in the tree. Maybe simply checking if the layer is already visible, and not turning off it is the case is enough to fix it?

https://github.com/compassinformatics/cpsi-mapview/blob/e35955ab9857185e22aad52ab57bd1053beba3d6/app/controller/LayerTreeController.js#L228-L233

geographika:
Can probably ignore now #583 is added?

geographika:
@JakobMiksch - unless there is anything you have concerns about, I will close this.

JakobMiksch:
@geographika from what I see do not have any concerns closing. However, I have not tested it again. Let me know if I shall test it locally again to confirm.

geographika:
Tested in child applications and all is ok. Closing.

PR #583, opened 2022-08-02
Remove setting of layer visibility from layer tree filter

geographika:
Quick fix for #581. This simply removes trying to handle setting layer visibility in the filter. 
The initial rationale was to handle the case where a user logs out and then logs back in after their user roles have been modified (without refreshing the browser). This use case may never happen, and even if it does, then changing which layers are visible in the application may be more confusing than helpful. 

Removing this logic keeps things simpler, and all that is lost is layers with visibility set to `true` by default have to be switched on manually in the case when logging in and out with different roles.


geographika:
@JakobMiksch - I'm merging for now so we can carry on testing in client apps. We can discuss next week if you've any thoughts on the above.

Issue #560, opened 2022-06-14
Add Tracing Functionality to Drawing Tools


geographika:
Design Notes
--------------

See the `LineString` example at https://openlayers.org/en/latest/examples/tracing.html

Allow a user to select part of an existing line in a vector layer to create a new partial line. 

A couple of options for the design:

- add to the new drawing/snapping tool at #556
- create as a mixin that could be added to any drawing tool

The tool should be configurable as to which layer(s) could be traced. This could be doneusing the `layerKey` property available to
all layers in the cpsi-mapview project. 

```
        tracingLayerKeys: ['CHANNELNETWORKVECTOR_WFS', 'NETWORKJUNCTIONS_WFS', 'NETWORKEDGES_WFS'],
```

Tests should be added/enhanced to ensure code coverage.

Questions
----------

- Is it possible to trace more than one layer?
- Is it possible for the tool to trace both lines and polygons?

geographika:
Note - a key requirement, which is not in the demo, is to allow tracing over connected line features to make a new single line feature. These features could be in separate layers, but if this is problematic we could ensure they are in the same layer.

JakobMiksch:
@geographika I built a prototype to test what is possible.

It seems that multiple layers and also tracing both line and polygon layers works.

Tracing connected lines seems to be the trickiest. I am currently investigating approaches how to make it possible.

One question regarding the topology of the line features: It seems the line features of the example line layer never cross each other and only touch each other at the start and/or the endpoint. See gif below.
Is this always true for the tracing layers?

![Peek 2022-08-09 12-43](https://user-images.githubusercontent.com/15704517/183629391-a74eb132-73c3-4e28-8aef-bef559857298.gif)

geographika:
@JakobMiksch I think we can ensure that any layers that require tracing won't have overlaps and  share a single vertex. Does this help with a solution?

JakobMiksch:
Yes, it does. It makes it easier to find the connected edges.

geographika:
> @geographika tracing over connected features worked for one layer so far. See gif below. But there are many edge cases that are not handled yet. More layers should be possible as well. However, my suggestion is to first built a working PR with only one referenced layer. Then we could discuss the next step. Is this okay for you?

That sounds perfect - love the GIF! I'll double check our side on how important multiple layers are for tracing.

geographika:
Implemented with #586 and #590. Edge cases and issues will be logged separately.

PR #585, opened 2022-08-10
Add 3 new edge vector layers and styles

geographika:
3 new line vector layers have been added to the demo MapServer (and deployed) via https://github.com/compassinformatics/mapview-demo/pull/7

@JakobMiksch - these layers are representative of some of the features we will be attempting to trace using #560. 

This pull request adds them to the layer tree, with basic SLD styles. It also moves all the vector testing layers to a subfolder. 

@JakobMiksch @jansule - let me know if it is ok to merge this, or if I should hold off while you are testing.

### DIFF
--- a/resources/data/layers/default.json
+++ b/resources/data/layers/default.json
@@ -390,6 +390,54 @@
     },
     {
       "layerType": "wfs",
+      "layerKey": "FOOTWAYS_WFS",
+      "featureType": "footways",
+      "noCluster": true,
+      "idProperty": "EdgeId",
+      "styles": [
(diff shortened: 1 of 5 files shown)


JakobMiksch:
@geographika I can not find the "Cycleways" on the map. (the other new layers worked)

Otherwise this can be merged from side.

geographika:
There are only a handful of cycleway features (unfortunate for residents of Tipperary!). E.g. the permalink at #map=14.107709462049973/-862309.85/6908883.73/0

PR #586, opened 2022-08-10
Add tracing to drawing tool

JakobMiksch:
- references #560 
- enables tracing via mixin
- tracing of multiple layers possible
- tracing of both lines and polygons possible
- tracing result is transferred via event to parent component
- enables tracing of connected edges that touch each other at startpoint or endpoint 
- tracing back to a previous features or tracing a branching edge can be done via hovering back

follow up tasks:
- [x] tests (basic tests added)
- [ ] continue tracing on a disconnected feature
- [ ] support edges that are connected by interior points and not only startpoint or endpoint - see https://github.com/compassinformatics/cpsi-mapview/pull/590 for an attempt to fix it


JakobMiksch:
This PR is a testable preview of the tracing functionality. see gif below.

- It currently supports only lines that touch at the start and/or endpoint (not interior point)
- by hovering over a previous line the tracing is reverted to that line
- it might still contain some unhandled corner-cases 
- for visual debugging the linestring where the user can trace on can be made visible, via a flag (https://github.com/compassinformatics/cpsi-mapview/pull/586/files#diff-97f2716461b3c003c313ca12b98590f4421986b1f7154e72ec51b2aabc2d50b8R197) 

@geographika what is your impression regarding usability so far?

Peek 2022-08-10 13-59

geographika:
@JakobMiksch - really nice! It is going to make my backend routing tools redundant!

I was very impressed when I traced one feature, then clicked on another unconnected edge and was able to start tracing again on the new edge. Not sure if this is by design?

It might be worth switching to the Service layer added in #585 (now merged) for testing/demo purposes as this is one of the datasets the tool will be initially used on. 

Just for info (you are probably aware), I got a few console errors after awhile:

image

app/controller/button/TracingMixin.js
```diff
@@ -0,0 +1,275 @@
...
+        var me = this;
+
+        if (me.getView()) {
+            me.getView().fireEvent('tracingstart');
+        }
+
+        // get tracing layers
+        me.tracingLayers = [];
+        Ext.each(tracingLayerKeys, function (key) {
+            var foundLayers = BasiGX.util.Layer.getLayersBy('layerKey', key);
```
jansule, line 44: Is there a helper variable (maybe in BasiGX) that contains the name of the identifying property? If so, it might be better to use that variable instead of the hardcoded value `layerKey`.

app/controller/button/TracingMixin.js
```diff
@@ -0,0 +1,275 @@
...
+            if (foundLayers.length === 1) {
```
jansule, line 45: Maybe it's more robust to check for `length > 0` and then still always use `foundLayers[0]`

app/controller/button/TracingMixin.js
```diff
@@ -0,0 +1,275 @@
...
+            geometry: new ol.geom.LineString([]),
+        });
+        me.tracingFeatureArray = [];
+        me.tracingStartPoint = null;
+        me.tracingEndPoint = null;
+
+
+        // For debugging the line where the user can trace to can be displayed.
+        var debug = false;
+        if (debug) {
```
jansule, line 68: I'm not sure, if we should keep the code as is. Maybe better provide a proper config variable to the mixin/component that can be switched to true/false.

app/controller/button/TracingMixin.js
```diff
@@ -0,0 +1,275 @@
...
+                        color: [255, 255, 0, 0.5],
+                        width: 25
+                    })
+                })
+            });
+            me.map.addLayer(tracingVector);
+        }
+
+        // the tracing util
+        me.util = CpsiMapview.util.Tracing;
```
jansule, line 84: How about `tracingUtil` instead?

app/controller/button/TracingMixin.js
```diff
@@ -0,0 +1,275 @@
...
+                    // find coordinate on found feature
+                    if (me.tracingFeatureArray.includes(foundFeature)) {
+                        coordOnFoundFeature = me.map.getCoordinateFromPixel(event.pixel);
+                    }
+
+                    // update tracing feature and array
+                    if (me.tracingFeatureArray.includes(foundFeature)) {
+
+                        // check if found feature is last of array
+                        var i = me.tracingFeatureArray.indexOf(foundFeature);
```
jansule, line 203: It might make sense to refactor this functionality into a dedicated function, in order to improve readability and testability.

app/controller/button/TracingMixin.js
```diff
@@ -0,0 +1,275 @@
...
+                        var tracingCoords = tracingGeom.getCoordinates();
+
+                        var touchingStartEnd = me.util.linesTouchAtStartEndPoint(foundGeom, tracingGeom);
+
+                        var tracingInteriorTouchesFoundFeatureStartEnd = me.util.lineInteriorTouchesLineStartEnd(tracingGeom, foundGeom);
+
+                        var tracingStartEndTouchesFoundInterior = me.util.lineStartEndTouchesLineInterior(tracingGeom, foundGeom);
+
+                        if (touchingStartEnd) {
+                            var coords = foundGeom.getCoordinates();
```
jansule, line 239: See comment above

app/util/Tracing.js
```diff
@@ -0,0 +1,331 @@
...
+            return ((a % b) + b) % b;
+        },
+
+        /**
+         * Checks if a LineString is not empty.
+         *
+         * @param {ol.Feature} feature The feature to check
+         * @returns {Boolean} If LineString is not empty
+         */
+        lineStringNotEmpty: function (feature) {
```
jansule, line 33: Minor: It might be easier to work with functions that return booleans, if the function name does not already include a negation, since these functions will probably be used with other logical operator in e.g. if-statements.

So a `lineStringEmpty` or a `lineStringPopulated` will be easier to work with, as we will not have a doubled negation.

Compare

```js
if (!lineStringEmpty) {
```

to

```js
if (!lineStringNotEmpty) {
```

app/util/Tracing.js
```diff
@@ -0,0 +1,331 @@
...
+         *
+         * @param {ol.coordinate.Coordinate[]} coordinateArray The coordinate array
+         * @param {ol.coordinate.Coordinate} coordToFind The coordinate to find
+         * @returns {Number} The index of the coordinate, or -1 if not found
+         */
+        getCoordIndex: function (coordinateArray, coordToFind) {
+            var found = Ext.Array.findBy(coordinateArray, function (c) {
+                return Ext.Array.equals(c, coordToFind);
+            });
+            return coordinateArray.indexOf(found);
```
jansule, line 60: If you use `Ext.Array.reduce` instead of `Ext.Array.findBy` you can omit the second loop.

app/util/Tracing.js
```diff
@@ -0,0 +1,331 @@
...
+        },
+
+        /**
+         * Checks if two LineString geometries are touching at only startpoint and/or endpoint.
```
jansule, line 64: Minor: Please add newline below function description

app/util/Tracing.js
```diff
@@ -0,0 +1,331 @@
...
+
+            var bFirst = bLineCoords[0] || [];
+            var bLast = bLineCoords[bLineCoords.length - 1] || [];
+
+            var lastFirst = Ext.Array.equals(aLast, bFirst);
+            var lastLast = Ext.Array.equals(aLast, bLast);
+            var firstFirst = Ext.Array.equals(aFirst, bFirst);
+            var firstLast = Ext.Array.equals(aFirst, bLast);
+
+            if (lastFirst) {
```
jansule, line 211: I'm not sure, if I like empty if blocks. Maybe it's cleaner to replace the last `else` with an `else if(!lastFirst)` or something alike

app/util/Tracing.js
```diff
@@ -0,0 +1,331 @@
...
+                return [];
+            }
+            // remove intersecting vertex
+            aLineCoords.pop();
+            var resultCoords = aLineCoords.concat(bLineCoords);
+            return resultCoords;
+        },
+
+        // TODO: this code was originally built for polygons and
+        //       might need more adjustment for simple linestrings
```
jansule, line 239: Is this still an open TODO?

jansule (commented):
Very nice @JakobMiksch. Just some minor nitpicks you might want to address

app/controller/button/TracingMixin.js
JakobMiksch: done with 1fe9356

app/controller/button/TracingMixin.js
JakobMiksch: done with 1d9075c

app/util/Tracing.js
JakobMiksch: good point! done with 2c936b1

app/util/Tracing.js
JakobMiksch: done with 5b791ae

app/controller/button/TracingMixin.js
JakobMiksch: changed to config with 476398f

app/util/Tracing.js
JakobMiksch: I moved this function to the follow-up PR #578  and therefore moved this note there

app/controller/button/TracingMixin.js
JakobMiksch: good point. I cannot find a constant defining the name of `layerKey` value. At other places in the code the layers are searched this way as well.
@geographika do you maybe know if something like this exists in CPSI?
otherwise this could be done in future.

app/controller/button/TracingMixin.js
JakobMiksch: done with 6708b7c79d11e54b6eb018c311a4784f7c40c933

app/controller/button/TracingMixin.js
JakobMiksch: done with 26aa2187efda7e1f9143e70fc3bd49397eee2451

app/util/Tracing.js
JakobMiksch: Actually yes, in most cases it works like expected, but there might be edge cases where it needs to be adapted.

app/util/Tracing.js
JakobMiksch: done with [62affc3](https://github.com/compassinformatics/cpsi-mapview/pull/586/commits/62affc3d231bf332821c2547844d05685aa2359f)

JakobMiksch:
@geographika I applied the review and did more refactoring, testing and cleanup.

JakobMiksch:
@geographika Please note this part: https://github.com/compassinformatics/cpsi-mapview/pull/586/files#diff-97f2716461b3c003c313ca12b98590f4421986b1f7154e72ec51b2aabc2d50b8R179-R185

The resulting coordinates are transferred to the parent component via an event. This could also be solved by directly applying the coordinates to the included `drawInteraction`. The current solution works, but changing it to another approach should be doable to do with reasonable effort.

app/controller/button/TracingMixin.js
geographika: We give all layers a `layerKey` property, and use this approach across the project for configuring and finding layers (set in default.json) so I think this is fine. I'm not sure what other variable available that could be used?

geographika (approved):
Can't wait to get this into a child application - please merge!

geographika:
> The current solution works, but changing it to another approach should be doable to do with reasonable effort.

Is there any particular advantage in this? Is it simpler cleaner to do it directly rather than through raising events and listeners?

geographika:
@JakobMiksch - as noted above, please merge when you are happy. I think there will be a few edge cases we can record when testing out the tools, and hopefully we can get some time to address these at a later stage. 

One possible key  issue is that sometimes it is impossible to fully trace a feature as the tracing "jumps" from one direction to another. For example, the line below can be traced up to this point:

image

If the mouse is moved slightly further then it draws it around the top, making it impossible to fully trace the feature:

image

app/controller/button/TracingMixin.js
JakobMiksch: Apparently the current approach has worked fine so far. I think this was just an idea for improvement. Maybe the name of the layerKey could be stored globally (or in the config file) , and there could be a function like `BasiGX.util.Layer.getLayersByKey(key)`. But this is just syntactic sugar.

JakobMiksch:
> > The current solution works, but changing it to another approach should be doable to do with reasonable effort.
> 
> Is there any particular advantage in this? Is it simpler cleaner to do it directly rather than through raising events and listeners?

The approach where the found coordinates are appended to the DrawInteraction directly, needs less code, but does the same.

JakobMiksch:
> @JakobMiksch - as noted above, please merge when you are happy. I think there will be a few edge cases we can record when testing out the tools, and hopefully we can get some time to address these at a later stage.
> 
> One possible key issue is that sometimes it is impossible to fully trace a feature as the tracing "jumps" from one direction to another. For example, the line below can be traced up to this point:
> 
> image
> 
> If the mouse is moved slightly further then it draws it around the top, making it impossible to fully trace the feature:
> 
> image

This edge case probably happens because the function `getPartialSegmentCoords()` was originally written for polygons and needs more adaptations for linestrings.

PR #588, opened 2022-08-10
upgrade BasiGX to v3.3.0

jansule:
title says it all...


geographika (approved):
Thanks for merging the https://epsg.io update

Issue #487, opened 2021-08-30
Add Icon to Coordinate Center Tool


geographika:
cpsi-mapview uses the `BasiGX.view.panel.CoordinateMousePositionPanel` in the bottom toolbar:

https://github.com/compassinformatics/cpsi-mapview/blob/7becdc33cb590a6546ef05e9986244d390c7b6a0/app/view/toolbar/MapFooter.js#L22

![image](https://user-images.githubusercontent.com/490840/131339067-eabfad0f-f7bf-4e1a-87fc-eb0a284533ff.png)

The old mapview UI included assitional functionality to place an icon at the centre of the map when using the coordinate tool. The icon could then be click to remove it from the map - see gif:

![Pavement Management System](https://user-images.githubusercontent.com/490840/131339289-3998b75c-ed58-490e-b137-8539619c10c8.gif)

The `BasiGX.view.panel.CoordinateMousePositionPanel` could be subclassed to add this functionlality (or added as an option to the BasiGX component?).

jansule:
Partially solved in https://github.com/terrestris/BasiGX/pull/699

jansule:
Fully solved with https://github.com/compassinformatics/cpsi-mapview/pull/589

PR #589, opened 2022-08-11
Enable showing marker on centerToCoordinate click

jansule:
Title says it all...

Waits for https://github.com/terrestris/BasiGX/pull/699 to be merged and published. BasiGX submodule can then be updated.

Solves https://github.com/compassinformatics/cpsi-mapview/issues/487

compass-marker-on-coordinate

### DIFF
--- a/app/view/toolbar/MapFooter.js
+++ b/app/view/toolbar/MapFooter.js
@@ -21,6 +21,16 @@ Ext.define('CpsiMapview.view.toolbar.MapFooter', {
             xtype: 'basigx-combo-scale'
         }, {
             xtype: 'basigx-panel-coordinatemouseposition',
+            showMarker: true,
+            markerStyle: new ol.style.Style({
+                text: new ol.style.Text({
+                    font: 'normal 2em "font-gis"',
+                    text: '\uea16',
+                    fill: new ol.style.Fill({
+                        color: 'red'
+                    })
+                })
+            }),
             epsgCodeArray: ['EPSG:4326', 'EPSG:29902', 'EPSG:2157'],
             activeEpsgCode: 'EPSG:29902',
             segmentedButtonLimit: 3


app/view/toolbar/MapFooter.js
```diff
@@ -21,6 +21,16 @@ Ext.define('CpsiMapview.view.toolbar.MapFooter', {
             xtype: 'basigx-combo-scale'
         }, {
             xtype: 'basigx-panel-coordinatemouseposition',
+            showMarker: true,
+            markerStyle: new ol.style.Style({
```
JakobMiksch, line 25: Meanwhile there are quite many predefined styles defined in the style util: https://github.com/compassinformatics/cpsi-mapview/blob/master/app/util/Style.js

Just wondering if it would make sense to store this style in the style util as well.

JakobMiksch (approved):
I tried it and it works well. 

One non-blocking thought inside.

app/view/toolbar/MapFooter.js
jansule: Good point! Will update accordingly

geographika:
@jansule - looks good. One coment in terms of functionality - would it be possible to leave the marker in place unless the marker itself is clicked on? If clicking on the marker it would remove the marker, and also prevent any other map actions from triggering (such as the default Feature Info tool).

jansule:
> @jansule - looks good. One coment in terms of functionality - would it be possible to leave the marker in place unless the marker itself is clicked on? If clicking on the marker it would remove the marker, and also prevent any other map actions from triggering (such as the default Feature Info tool).

Yes. I can update the logic accordingly

jansule:
Partially solved in https://github.com/terrestris/BasiGX/pull/700

jansule:
@geographika  I changed event handling accordingly. Now only waits for a new release of BasiGX

compass-marker-click-event

geographika (approved):

geographika:
@jansule - I tested with the latest changes and your [jansule:coordinate-marker-fix](https://github.com/jansule/BasiGX/tree/coordinate-marker-fix) branch - all works great. When the BasiGX project is updated please merge this. Thanks!

PR #590, opened 2022-08-11
Tracing Add-On: Support edges touching at interior points

JakobMiksch:
- references #560 
- This PR adds the attempt to also include edges that touch each other at the interior points.

#### TODO:
- [ ] tracing consecutively on many features does not work
- [ ] probably many edge cases not discovered yet

### DIFF
--- a/app/controller/button/DrawingButtonController.js
+++ b/app/controller/button/DrawingButtonController.js
@@ -681,6 +681,8 @@ Ext.define('CpsiMapview.controller.button.DrawingButtonController', {
         if (me.drawLayer) {
             me.map.removeLayer(me.drawLayer);
         }
+
+        me.cleanupTracing();
     },
 
     init: function () {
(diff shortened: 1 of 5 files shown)


JakobMiksch:
@geographika I did refactoring and cleanup. Also pulled the latest changes from #586  and master

geographika:
@JakobMiksch - I think it would be good to merge this too so we can carry out full testing and log any issues/edge cases to prioritise addressing in a later phase.

JakobMiksch:
@geographika The PR still has many edge cases. One missing feature is that tracing consecutive lines and hovering back to previous lines, does not work yet. But this should only affect cases where edges touch each other at the interior point. Feel free to merge, if this is fine for you.

geographika:
I'm going to merge, so we can assess which edge cases are a priority. Thanks @JakobMiksch.

PR #596, opened 2022-08-22
Split out part of onRowDblClick in cmv_grid to allow for custom logic when determining associatedEditWindow and record Id 

fergaldoyle:
In cases where a grid has mixed entity types, and in cases where Ids in a mixed grid may be negated to avoid Id conflicts, this PR  spits out some of the logic from `onRowDblClick` into `getRecordEditConfig` which is intended to be overridden in a subclass if needed.

Example usage:
```javascript
Ext.define('Pms.view.grids.CombinedEdgesGridController', {
    extend: 'CpsiMapview.controller.grid.Grid',
    alias: 'controller.pms_combinededgesgrid',
    requires: [
        'CpsiMapview.controller.grid.Grid'
    ],
    // override the method in the parent class to apply custom logic
    // in this mixed grid
    getRecordEditConfig: function (vm, record) {
        var id = record.getId();
        return {
            associatedEditWindow: vm.get('associatedEditWindow'),
            associatedEditModel: vm.get('associatedEditModel'),
            id: Math.abs(id)
        };
    },
});
```

### DIFF
--- a/app/controller/grid/Grid.js
+++ b/app/controller/grid/Grid.js
@@ -313,6 +313,21 @@ Ext.define('CpsiMapview.controller.grid.Grid', {
         me.updateAssociatedLayers(force);
     },
 
+    /**
+    * Gets associatedEditWindow, associatedEditModel and record Id
+    * from a ViewModel and record Model. It is a separate function
+    * so that it might be overridden in a subclass to return custom values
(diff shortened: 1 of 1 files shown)


geographika:
Thanks @fergaldoyle !

PR #599, opened 2022-08-30
Allow custom logic in clickableLayerConfigs

fergaldoyle:
This PR allows `clickableLayerConfigs` accept both plain objects and a function which returns a config object.  The function receives the clicked feature as an argument which can be then used in custom logic. E.g. check a property on the feature and open the relevant edit window. The function returns an Id instead of keyField which allows for pre-processing of the Id before it is used by the model load function.

```javascript
{
  layerKeyName: function (feat) {
    return {
      id: Math.abs(feat.get('ObjectId')), // the unique feature Id to open in a form
      modelClass: 'CpsiMapview.model.ModelName', // the name of the class used to load a model
      editWindowClass: 'CpsiMapview.view.EditWindow' // the name of the form view class to open when clicking on a feature
    }
  }
}
```

### DIFF
--- a/app/controller/MapController.js
+++ b/app/controller/MapController.js
@@ -27,6 +27,19 @@ Ext.define('CpsiMapview.controller.MapController', {
     *   }
     * }
     *
+    * You also use a function to return a configuration object which returns an Id instead of a
+    * keyField which allows custom logic to be applied. The function receives a single argument which
+    * is the clicked ol.Feature
+    * {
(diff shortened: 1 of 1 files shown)


geographika:
@fergaldoyle - just to clarify - the use case for this is to allow an application to use a different key/id/custom value to pass to a service, or depending on the feature properties use different forms and models?

fergaldoyle:
@geographika That's correct.

PR #605, opened 2022-09-23
Fix class names filenames mis-match

fergaldoyle:
A couple of classes / files had a naming mis-match which caused tests in projects using cpsi-mapview to fail in certain circumstances.

The class name / filename mis-match caused Ext.require to look for js files that didn't exist, causing certain Ext.require statments in tests to never resolve, meaning objects the tests were run against were undefined.

The class names changed look unlikely to be used in any project which consumes the cpsi-mapview project


geographika:
Thanks @fergaldoyle !

PR #608, opened 2022-09-29
Add option to only snap to visible layers/features

fergaldoyle:
Currently when snapping is set up in `cmv_drawing_button` and one of the configured snapping Layers is turned off, the snap interaction will still snap to the features of the invisible layer.

This PR adds option `allowSnapToHiddenFeatures` which defaults to `true` (current behavior)

If set to `false`, the snapCollection collection will be updated accordingly depending on layer visibility.

### DIFF
--- a/app/view/button/DrawingButton.js
+++ b/app/view/button/DrawingButton.js
@@ -129,9 +129,9 @@ Ext.define('CpsiMapview.view.tool.DrawingButton', {
         /**
          * Allows consumer of component choice of snapping to only visible layer features
          * defined in snappingLayerKeys, or snapping to layer features even if they
-         * are invisible. Defaults to true for backward compatibility.
+         * are invisible.
          */
-        allowSnapToHiddenFeatures: true
+        allowSnapToHiddenFeatures: false
     },
 
     /**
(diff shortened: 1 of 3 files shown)


app/view/button/DrawingButton.js
```diff
@@ -124,7 +124,14 @@ Ext.define('CpsiMapview.view.tool.DrawingButton', {
...
          */
-        drawInteractionSnapTolerance: 2
+        drawInteractionSnapTolerance: 2,
+
+        /**
+         * Allows consumer of component choice of snapping to only visible layer features
+         * defined in snappingLayerKeys, or snapping to layer features even if they
+         * are invisible. Defaults to true for backward compatibility.
+         */
+        allowSnapToHiddenFeatures: true
```
geographika, line 134: The snapping features haven't yet been released in any production app, so I think it would make sense to set this to `false` by default?

geographika (approved):

geographika:
@fergaldoyle would it be possible to add a test to run the `setSnapInteraction` function and add to https://github.com/compassinformatics/cpsi-mapview/blob/master/test/spec/controller/button/DrawingButtonController.spec.js ?

PR #609, opened 2022-10-06
Adding snapping tests and improve handling of unique features when re…

fergaldoyle:
…moving from the snapCollection


geographika:
Excellent work - thanks @fergaldoyle !

PR #612, opened 2022-11-02
Set tracingStartPoint and tracingEndPoint to the current snapped coord

fergaldoyle:
The current tracing functionality finds the coords on the tracing line closest to the coords clicked by the user. When the user is snapped to a vertex, the coords of the snapped vertex and the position of the curser will likely not be exactly the same. This leads to issues where the drawn start point will snap to the vertex, but the tracing line will start offset from the vertex, and can cause an overlap of the traced line which is not a simple geom which is causing issues. See example in screencast below. The curser if offset from the vertex, but snapping is holding the point on the vertex. The further out you are zoomed, the greater the overlap of the tracing.

https://user-images.githubusercontent.com/1889044/199502468-2cf21e88-011b-4c74-a8e6-510f63e39c08.mp4

This PR tracks the coords of snapped vertex/edge/nodes or clicks away from features, and those coords are then used to set the tracing start and end points instead of the cursor position, e.g:

https://user-images.githubusercontent.com/1889044/199505590-f10a1057-e7dc-4365-9d46-6415fe9ffb2d.mp4

### DIFF
--- a/app/controller/button/TracingMixin.js
+++ b/app/controller/button/TracingMixin.js
@@ -25,6 +25,16 @@ Ext.define('CpsiMapview.controller.button.TracingMixin', {
     }),
 
     /**
+     * Used to store the position of the last mouse coord, taking into account snapping
+     */
+    lastSnappedCoord: null,
+
+    /**
+     * Interaction to track and populate lastSnappedCoord
+     */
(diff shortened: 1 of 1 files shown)


geographika:
Makes sense from the videos. Thanks @fergaldoyle

PR #613, opened 2022-11-07
Allow merging to both start and end of existing line

fergaldoyle:
Currently with the `cmv_drawing_button`, you can only append to the last point of a previously drawn line. If you try to append to the start point, it will create a new line:

https://user-images.githubusercontent.com/1889044/200355790-b08c6c50-6bf0-4795-a378-19a59338078e.mp4

With this PR, you can append to either start or end of the previous line. If appending to the start, the coords of the previous line are reversed so the we end up with a single line with one start and one end point.

https://user-images.githubusercontent.com/1889044/200356769-14567c54-d1ea-41f7-a4e2-6e14624e1ca7.mp4


geographika:
Thanks @fergaldoyle !

PR #614, opened 2022-11-10
Add option to only show FeatureInfoWindow if features are found

fergaldoyle:
This PR adds the ability to set option `showFeatureInfoWindowOnlyIfContent` which will change the behavior of the FeatureInfoWindow  popup. 

Instead of the window always opening, then possibly being empty if no features are found, the window will only open if features are found in one of the layers enabled when the user clicked the map.

### DIFF
--- a/app/plugin/FeatureInfoWindow.js
+++ b/app/plugin/FeatureInfoWindow.js
@@ -46,7 +46,7 @@ Ext.define('CpsiMapview.plugin.FeatureInfoWindow', {
 
         cmp.on('cmv-mapclick', function (clickedFeatures, evt) {
             if (me.map.get('defaultClickEnabled')) {
-                me.requestFeatureInfos(clickedFeatures, evt);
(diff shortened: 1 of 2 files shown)


geographika:
Is there any use case where showing the Window with no features should exist? Wondering if `showFeatureInfoWindowOnlyIfContent` if required, and simply make this workflow the default.

fergaldoyle:
Probably not a use case to show the window if no features exist. I added the behaviour change as optional in case any project wanted to keep the legacy behaviour.

Can make it the default behavior if you think it's worthwhile

geographika:
@fergaldoyle - yes, I think making it the default behaviour would be simplest. It would be nice to raise a custom event if no features are found - an application could then hook in to display a `Ext.toast` or similar.

fergaldoyle:
@geographika Changes made

geographika:
For reference, the two new custom events that can be used to give feedback in the UI are:

- `cmv-featureinfowindow-no-features-found` - no features found in active layers
- `cmv-featureinfowindow-no-layers-enabled` - no layers to query

PR #627, opened 2023-01-05
Order column selection dropdown alphabetically, add test

fergaldoyle:
- Order the items in column selection menu in the data grid alphabetically
- Added a test

### DIFF
--- a/app/view/grid/Grid.js
+++ b/app/view/grid/Grid.js
@@ -237,12 +237,5 @@ Ext.define('CpsiMapview.view.grid.Grid', {
                 boxLabel: 'Page Records?',
                 handler: 'togglePaging'
             }]
-    }],
-
-    constructor: function() {
-        this.callParent(arguments);
-
-        //console.log('constructor', this)
-
-    }
+    }]
 });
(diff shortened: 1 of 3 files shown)


app/view/grid/Grid.js
```diff
@@ -236,5 +237,12 @@ Ext.define('CpsiMapview.view.grid.Grid', {
                 boxLabel: 'Page Records?',
                 handler: 'togglePaging'
             }]
-    }]
+    }],
+
+    constructor: function() {
```
geographika, line 242: Out of interest why is this required? Shouldn't the default constructor be called anyway?

test/spec/controller/grid/Grid.spec.js
```diff
@@ -9,4 +9,54 @@ describe('CpsiMapview.controller.grid.Grid', function() {
...
+
+            // get a reference to the menu by taking it from the spy arguments
+            var menu = onHeaderMenuCreateSpy.getCalls()[0].args[1];
+            var menuItems = menu.down('[itemId=columnItem]').menu.items.items;
+            var orderedTitles = menuItems.map(function (item) {
+                return item.text;
+            })
+
+            // deep array comparison - original order without sorting would be [ 'Test1', 'Test3', 'Test2' ]
+            expect(orderedTitles).to.eql(['Test1', 'Test2', 'Test3']);
```
geographika, line 55: Very nice!

app/view/grid/Grid.js
fergaldoyle: That's just a left over from debugging. I've removed it now in the latest commit.

geographika (approved):

PR #630, opened 2023-01-11
Order column selection dropdown alphabetically on all grids

fergaldoyle:
Using an Ext.menu.Menu override, detects if the menu is 'columnItem' and applies ordering if it is.


fergaldoyle:
Updated to use mixin instead of override

app/view/grid/Grid.js
```diff
@@ -15,6 +15,8 @@ Ext.define('CpsiMapview.view.grid.Grid', {
         'CpsiMapview.view.button.FeatureSelectionButton'
     ],
 
+    mixins: ['CpsiMapview.util.ColumnMenuOrderMixin'],
```
geographika, line 18: One question on this approach - if a grid is a subclass of `CpsiMapview.view.grid.Grid` and defines its own `mixins` property does it override the `['CpsiMapview.util.ColumnMenuOrderMixin']` or append to it? Not sure if we have any grids where this is the case at the moment

fergaldoyle: Mixins get inherited like other members so in this case, mixins defined in the subclass would exist along side `CpsiMapview.util.ColumnMenuOrderMixin`

fergaldoyle:
There's also the `Filters` menu item which appears under the `Columns` menu item. The items in this submenu are usually defined by a store or by hard coded values, both of which can are defining the order of the items. Often with items in a store the items are sorted by Id and not by Name.

Do we want the items in the `Filters` submenu to be ordered alphabetically like the `Columns` submenu? If so:
- Should that ordering be added to this PR mixin (and rename the mixin to something more generic) or should it be a separate mixin?  (The `Filters` item is added at some point after the header menu is open, so hooking into it is a bit more difficult compared to how we hook into the `Columns` item)
- Should there be some kind of flag (if possible) to prevent ordering of `Filters` submenu if the order of the store / hard coded items is preferred for some reason for a particular column?

geographika:
> There's also the `Filters` menu item which appears under the `Columns` menu item. The items in this submenu are usually defined by a store or by hard coded values, both of which are defining the order of the items. Often with items in a store the items are sorted by Id and not by Name.

In my view ordering should be based on the store order. These can be changed if/when required. There are several stores which are sorted so "legacy" items are moved the the bottom. 
Maybe revisit in a separate pull request if it becomes a requirement? Thoughts @select-8 ?

PR #631, opened 2023-01-13
Dynamically set the PROPERTYNAME param depending on style loaded and …

### LINKED ISSUE #628: Request minimum required fields for WFS Layers
An example WFS layer configuration (used in `default.json`) is shown below:

```
  {
    "layerType": "wfs",
    "layerKey": "MYLAYER_WFS",
    "geometryProperty": "msGeometry",
    "featureType": "MyLayer",
    "serverOptions": {
      "propertyname": "Field1,Field2,Field3
    },
    "styles": [
      {
        "name": "style1.xml",
        "label": "style1_Label.xml",
        "title": "Style 1"
      },
      {
        "name": "style2.xml",
        "label": "style2_Label.xml",
        "title": "Style 2"
      }
     ],
    "tooltipsConfig": [
      {
        "alias": "Field1",
        "property": "Field1"
      },
      {
        "alias": "Field2",
        "property": "Field2"
      }
     ]
```

This will request the WFS layer with a `PROPERTYNAME` parameter containing the list of fields in `serverOptions` > `propertyname` e.g. "Field1,Field2,Field3". The backend (MapServer) can then limit the fields it requests from a database view to just these fields. If `serverOptions` > `propertyname` is not set then all fields are requested. Only requesting the exact list of fields required can sometimes make a noticeable difference in download speed. 

Currently `serverOptions` > `propertyname` needs to contain _all_ the fields used by the tooltips configuration _and_ all the fields used in all the SLD styles configured for the layer. 

In layers with many SLD styles (e.g. ~10), then all WFS requests include all 10 fields, even though typically only 1 fields is actually required for the client-side styling. 

image

Switching styles causes a new server WFS request (although if all fields are present I'm not sure if/why this is required). 

Two possible optimistations seem possible:

- instead of setting `serverOptions` > `propertyname` "manually" in the config, build the exact list of required fields. This would include all `property` values from `tooltipsConfig` objects, and all properties required for the selected style. The style object would need to be modified to list these fields e.g.

```
      {
        "name": "style1.xml",
        "label": "style1_Label.xml",
        "properties": "Field1,Field3,LabelField"
        "title": "Style 1"
      },
```

Properties would need to be a comma-separated string as an SLD could make use of several properties for styling, and the associated `label` SLD file could also include additional fields. 

A backend scrript (Python) could parse the SLD files and create these fields lists automatically by searching for property names. 

The JS would then check the requested style and concatenate the tooltips and style properties, remove any duplicates, and set the `propertyname` to send as the WFS request. 

A second possible optimisation is when switching styles to check if the fields used for the newly selected style are already present in the OL layer. If the values are already available client-side then there should be no requirement to make a new server request.

fergaldoyle:
…tooltip config

resolves #628 

If `serverOptions.propertyname` has not been set in the config, the `PROPERTYNAME` is not set and all fields will be returned. Style switching and tooltip config doesn't change this because all properties are already available

On WFS layer creation, the values in `serverOptions.propertyname` and property names referenced in `tooltipConfig` are combined and sent in the `PROPERTYNAME` param. This means you can define only the properties used in the default style `serverOptions.propertyname` and let the `tooltipConfig` populate the rest

On WFS default style xml file load, or on switching styles, property names will be extracted from the SLD XML. If there are property names found in the SLD which have not already been requested, these property names will be added to the list and the source refreshed (a new WFS request with required fields for the current style and `tooltipConfig`)

The property names originally defined in `serverOptions.propertyname` config will always be sent along side the dynamically generated property names. This is to ensure that certain properties that might not be in a style or a `tooltipConfig` are always available so that functionality that this PR (https://github.com/compassinformatics/cpsi-mapview/pull/599) enables will continue to work

Generated and mergerd property names are cleaned and duplicates removed. The are treated as case sensitive.

### DIFF
--- a/app/factory/Layer.js
+++ b/app/factory/Layer.js
@@ -422,7 +422,6 @@ Ext.define('CpsiMapview.factory.Layer', {
      * @return {ol.layer.Vector}  WFS layer
      */
     createWfs: function (layerConf) {
-        var url = layerConf.url;
         var baseUrl = layerConf.url;
         // transform OL2 properties to current ones supported by OL >=v3
         var olSourceProps = this.ol2PropsToOlSourceProps(layerConf.openLayers);
@@ -447,6 +446,7 @@ Ext.define('CpsiMapview.factory.Layer', {
             SRSNAME: srid
         };
 
+        // merge serverOptions into params
         Ext.iterate(serverOptions, function (key, val) {
             key = (key + '').toUpperCase();
             if (key in params && val === null) {
@@ -456,8 +456,6 @@ Ext.define('CpsiMapview.factory.Layer', {
             }
         });
 
-        url = Ext.String.urlAppend(url, Ext.Object.toQueryString(params));
-
         var olSourceConf = {
             format: new ol.format.GeoJSON(),
             strategy: ol.loadingstrategy.bbox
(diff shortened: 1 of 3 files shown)


geographika:
Very nice @fergaldoyle, especially the parsing of the SLD files to get the field names. It makes the PL vector layers much faster being able to avoid loading all fields required for all styles. 

One issue (which appears to have been an existing bug, but highlighted by this change) relates to the switch layers (these are single layers in the layer tree but which switch from WMS to WFS when zoomed in). 

There is a race condition when zooming to and switching to WFS, and the vector layer is loaded twice - once without the fields required for the style (and if this finishes last it results in an empty layer as there is no field to symbolise on), and once when the SLD file is loaded.

SLD load call:

https://github.com/compassinformatics/cpsi-mapview/blob/f7310451c24e29a9d6248f57a20ba36810beb311/app/util/SwitchLayer.js#L157

Adding layer to a collection - this also triggers the vector load (but without fields required for the SLD):

https://github.com/compassinformatics/cpsi-mapview/blob/f7310451c24e29a9d6248f57a20ba36810beb311/app/util/SwitchLayer.js#L173-L174

The removal/addition of a layer is a workaround to prevent a layer appearing multiple times in the tree.
Even if it were possible to use `layerCollection.setAt(0, newLayer);` the vector layer is still loaded. 

Any ideas for a solution to avoid the double-load in this case? Maybe along the lines of the `loadSld` being a promise and only adding the layer (in the case of vector layers) in this case?
Supressing `ol.collection` events seems a riskier approach.

fergaldoyle:
@geographika Added an update to handle the above. Aborts an in progress xhr request for the source if new fields are found in the style xml before refreshing the source creating a new xhr request

PR #633, opened 2023-01-24
Set qtip to fix issue where tooltip is not updated

### LINKED ISSUE #629: Tooltips not showing layer ranges
The TreeColumnInResolutionRange plugin no longer seems to show the min and max ranges for a layer:

https://github.com/compassinformatics/cpsi-mapview/blob/master/app/plugin/TreeColumnInResolutionRange.js

See the demo https://compassinformatics.github.io/cpsi-mapview/

image

This was previously working when introduced in #123

Not sure where the `description` value in `node.getOlLayer().get('description');` is set.

fergaldoyle:
Fixes #629 

Everything worked as expected up until `node.triggerUIUpdate()`. The UI/tooltips didn't update as they did before, maybe a change in ExtJS itself?

Setting the `qtip` instead updates the tooltip as expected.

### DIFF
--- a/app/plugin/TreeColumnInResolutionRange.js
+++ b/app/plugin/TreeColumnInResolutionRange.js
@@ -74,7 +74,7 @@ Ext.define('CpsiMapview.plugin.TreeColumnInResolutionRange', {
             var changed = description !== descriptionBefore;
             node.getOlLayer().set('description', description);
             if (changed) {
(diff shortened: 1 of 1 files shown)


geographika:
Thanks @fergaldoyle. Yes it could be something in ExtJS - the plugin would have been written for ExtJS 6.2 originally.

PR #636, opened 2023-03-06
Add Group Edit Mixin

geographika:
This pull request allows a new drop-down menu to be added to grid columns to allow them to be "group edited":

image

image

When the "Group Edit" mode is selected, a checkbox column appears on the left of the grid. A user can then select records, and update multiple records by selecting a value from the dropdown list of the column:

image

image

image

An example column configuration is shown below:

            {
                text: '<i class="x-fa fa-pencil"></i> Editable Example',
                dataIndex: 'exampleField',
                flex: 0.3,
                hidden: false,
                xtype: 'booleancolumn',
                trueText: 'Yes',
                falseText: 'No',
                groupEditable: true,
                groupEditService: 'exampleServiceUrl',
                groupEditDataProp: 'myExampleField',
                filter: {
                    type: 'boolean',
                },
            },

A service call is made using the following format, `idFieldName` is the `idProperty` used for the grid. 

`{idFieldName: [260891, 522021, 522020, 521148], myExampleField: 1}`

### DIFF
--- a/app/controller/grid/GroupEditMixin.js
+++ b/app/controller/grid/GroupEditMixin.js
@@ -1,0 +1,209 @@
+Ext.define('CpsiMapview.controller.grid.GroupEditMixin', {
+    extend: 'Ext.Mixin',
+    requires: ['Ext.menu.Menu'],
+    mixinConfig: {
+        on: {
+            init: function () {
+                var me = this;
+                var view = this.getView();
+
+                view.on('beforerender', function (grid) {
+                    var dropdownMenu = grid.headerCt.getMenu();
+                    dropdownMenu.on({
+                        beforeshow: me.onHeaderMenuBeforeShow,
+                        scope: me
+                    });
+
+                    // Add custom menu items to the default grid menu
+                    dropdownMenu.insert(dropdownMenu.items.length - 2, [{
+                        itemId: 'groupEditorMenuItem',
+                        text: 'Group Edit',
+                        tooltip: 'Group Edit mode must be enabled to use this menu. If you do not see the Group Edit button, you may not have sufficient permissions for grid editing.',
+                        bind: {
+                            disabled: '{!isGroupEditingEnabled}'
+                        }
+                    }]);
+                });
+
+                // Hide the checkbox selection column on initial load
+                view.on('render', function (grid) {
+                    var c = grid.columnManager.getFirst();
+                    if (c) c.hide();
+                });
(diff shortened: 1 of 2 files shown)


geographika:
Update notes on how to integrate this into a grid.

1. Create a custom controller with the mixin (it is not in the `CpsiMapview.controller.grid.Grid` by default - although perhaps could be). It does not appear possible to configure a grid type with mixins when declaring a view, so extending the `CpsiMapview.controller.grid.Grid` seems to be the only way to add the mixin:

```
// custom controller
Ext.define('MyProject.view.grids.MyGridController', {
    extend: 'CpsiMapview.controller.grid.Grid',
    alias: 'controller.myproject_mygrid',
    requires: [
        'CpsiMapview.controller.grid.Grid'
    ],
    mixins: [
        'CpsiMapview.controller.grid.GroupEditMixin'
    ]
});
```

2. In the view model for the grid some additional properties are required:

```js
Ext.define('MyProject.view.grids.MyGridModel', {
    extend: 'CpsiMapview.model.grid.Grid',
    alias: 'viewmodel.myproject_mygrid',
    data: {
        ...
        // for group editing
        idProperty: 'myIds', // property containing Ids to send to the service
        serviceUrl: '/WebServices/myservice/update/', // the URL of the update service
    },
    formulas: {
        // this logic is required to show the "Group Edit" button in the grid
        isGroupEditingVisible: function () {
            // can create custom logic based on logged in user or simply return true
            return true;
        }
    }
```

3. In the grid itself we can add a selection model with checkboxes:

```js
Ext.define('MyProject.view.grids.MyGrid', {
    extend: 'CpsiMapview.view.grid.Grid',
    requires: [
        ...
        'MyProject.view.grids.MyGridController', // custom controller defined above
        'GeoExt.selection.FeatureCheckboxModel',
    ],
    xtype: 'myproject_mygrid',
    viewModel: 'myproject_mygrid',
    controller: myproject_mygrid', // custom controller

    // Added for Group Edit
    selModel: {
        selType: 'featurecheckboxmodel',
        checkOnly: false,
        selectStyle: new ol.style.Style({
            stroke: new ol.style.Stroke({
                width: 10,
                color: '#0ff'
            })
        })
    },
```

Once configured calling the service will send the following JSON payload, where value is the value selected in the groupedit menu. 

```json
{"myIds":[-1129,-1128],"value":11}      
```

PR #637, opened 2023-03-22
Add ability to override default.json and tree.json paths

fergaldoyle:
This PR allows projects using `cpsi-mapview` to override the location of default.json and tree.json (layerConfig and treeConfig).
By default, current behaviour is maintained. i.e. mapview will load `resources/data/layers/default.json` and `resources/data/layers/tree.json`  unless the project overrides `resourcePathsDeferred`, in which case it can resolve/set any values it needs to.

`Ext.Deferred` is used so that the project can resolve custom paths at some point in future, e.g. after the login event which is fired after the map is initialised.

### DIFF
--- a/app/controller/LayerTreeController.js
+++ b/app/controller/LayerTreeController.js
@@ -207,17 +207,20 @@ Ext.define('CpsiMapview.controller.LayerTreeController', {
      * @return {Ext.Promise} Promise resolving once the JSON is loaded
      */
     loadTreeStructure: function () {
-        return new Ext.Promise(function (resolve, reject) {
-            Ext.Ajax.request({
-                url: 'resources/data/layers/tree.json',
-                method: 'GET',
-                success: function (response) {
-                    var respJson = Ext.decode(response.responseText);
-                    resolve(respJson);
-                },
-                failure: function (response) {
-                    reject(response.status);
-                }
+        var app = Ext.getApplication ? Ext.getApplication() : Ext.app.Application.instance;
+        return app.resourcePathsDeferred.then(function (resourcePaths) {
+            return new Ext.Promise(function (resolve, reject) {
+                Ext.Ajax.request({
+                    url: resourcePaths.treeConfig,
+                    method: 'GET',
+                    success: function (response) {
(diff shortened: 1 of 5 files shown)


geographika:
Looks good @fergaldoyle.
Could you give an example of how to set `resourcePathsDeferred` to return different JSON files? I presume you need to set a function here that returns a config object of the form below?

```
{
      layerConfig: 'resources/data/layers/custom.default.json',
      treeConfig: 'resources/data/layers/custom.tree.json'
 }
```

fergaldoyle:
@geographika I might switch to using a normal Promise since Deferreds are out of date.

fergaldoyle:
I updated to use Promises before I saw your reply.  I changed from using `resourcePathsDeferred` to a function `getResourcePaths` which returns a promise, and that function can be overridden by a project like PMS, e.g:

```javascript
init: function () {
    var me = this;

    var getResourcePathsPromise = new Ext.Promise(function (resolve) {
        Ext.GlobalEvents.on('login', function () {
          // base paths
          var config = {
              layerConfig: 'resources/data/layers/',
              treeConfig: 'resources/data/layers/'
          };
          
          // file names for all other roles
          var defaultJsonFile = 'default.json';
          var treeJsonFile = 'tree.json';

          // override file names for different roles
          if (IS_ROLE_A) {
              defaultJsonFile = 'role_a_default.json';
              treeJsonFile = 'role_a_tree.json';
          } else if (IS_ROLE_B) {
              defaultJsonFile = 'role_b_default.json';
              treeJsonFile = 'role_b_tree.json';
          }

          config.layerConfig += defaultJsonFile;
          config.treeConfig += treeJsonFile;
          resolve(config);
        });
    });

    me.getResourcePaths = function () {
        return getResourcePathsPromise;
    }

    me.callParent();
}
```

Deferreds originated in libraries like jQuery and Ext, but never made it to the native JS spec since they are considered an anti-pattern.

geographika:
@fergaldoyle - the updated approach and example seem much simpler - thanks. Merging now.

Issue #486, opened 2021-08-30
Karma test runner issues


geographika:
This is a placeholder issue with notes onto investigations into occasional failing tests, and console errors when running the tests in a browser. Related to #475. 

Example of [test error](https://github.com/compassinformatics/cpsi-mapview/runs/3472884711#step:11:208) and[105 tests run](https://github.com/compassinformatics/cpsi-mapview/runs/3472884711#step:11:214) when there are 113..!

The addition of coveralls highlighted an issue where the full tests suite was not always fully executed (but tests didn't fail), and so coverage dropped even though more tests may have been added. 

To view errors in the browser, navigate to root of cpsi-mapview project and run:

`karma start karma.conf.js --single-run`

Note there are also similar errors when running the GeoExt test suite (on which the karma configs in cpsi-mapview are based). However the GeoExt suite always seems to fully run. 

For GeoExt use:

`karma start ./test/karma.conf.js --browsers Chrome --single-run=False --debug --auto-watch`

The console errors can be viewed in Chrome debugger:

![image](https://user-images.githubusercontent.com/490840/131335027-24d33170-612a-4b57-80e3-4fdbe57d8844.png)

### File Loading Notes

The source JS files are listed in `karma-conf.common.js`. They can be set to have an `include` property set to `true`

https://github.com/compassinformatics/cpsi-mapview/blob/7becdc33cb590a6546ef05e9986244d390c7b6a0/karma-conf.common.js#L16

When this is set the JS files are added as script tags to the page generated by karma:

![image](https://user-images.githubusercontent.com/490840/131335132-af9bfecd-7ef0-48ef-a339-6c32b75bc8b9.png)

In addition to these the `Ext.Loader` is configured to load ExtJS JS files on request (see https://github.com/compassinformatics/cpsi-mapview/blob/master/test/load-tests.js):

```
    Ext.Loader.setConfig({
        enabled: true,
        paths: {
            'GeoExt': '../lib/geoext3',
            'BasiGX': '../lib/BasiGX',
            'CpsiMapview': '../app'
        }
    });
```

However if all files are already included why is the loader needed?
Also are the following required in the tests - `Ext.Loader.syncRequire(['GeoExt.toolbar.WfsPaging']);` (see for example [here](https://github.com/geoext/geoext/blob/9956ce66420c23c5d80a471361474759c77b8987/test/spec/GeoExt/toolbar/WfsPaging.test.js#L1)

[...]

PR #643, opened 2023-06-19
Add showGridWindow titleOverride option

fergaldoyle:
This PR allows a grid window title to be set to a custom value in the scenario where the layer doesn't exist in tree.json (currently where the window title is retrieved from)

Usage:
```
this.showGridWindow(layerKey, 'My Custom Title')
```

### DIFF
--- a/app/util/MenuMixin.js
+++ b/app/util/MenuMixin.js
@@ -7,7 +7,7 @@
 
     mixins: ['CpsiMapview.util.EditWindowOpenerMixin'],
 
-    showGridWindow: function (layerKey) {
+    showGridWindow: function (layerKey, titleOverride) {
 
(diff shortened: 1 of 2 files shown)


geographika:
I don't think it would apply in this case, but with grid windows any viewModel data properties get copied to the parent window so would be accessible there too: https://github.com/compassinformatics/cpsi-mapview/blob/6afab37bdd0eba83ba2b2714516d55123b7f0c0e/app/util/Grid.js#L102

geographika:
Also I needed to rerun the CI as the first time it failed: https://github.com/compassinformatics/cpsi-mapview/actions/runs/5319171837/jobs/9631372533
Tests are still somewhat flaky. See also https://github.com/compassinformatics/cpsi-mapview/issues/486

PR #644, opened 2023-07-03
Move padlock hidden binding to a formula so it can be overridden

fergaldoyle:
This PR allows individual form windows hide the padlock icon even if `Application.enableIsLocked` is true. 

To hide the padlock icon for a particular window, set create formula `isPadlockVisible` and return false


geographika:
Just to note https://docs.sencha.com/extjs/7.7.0/classic/Ext.data.Model.html#method-isPhantom is listed as a "private" method, but simply returns the `data.phantom` property. 
You have the check for the `isPhantom` method anyway so it should be fine.

PR #647, opened 2023-07-21
Npm prep

fergaldoyle:
This PR configures `package.json` ready to publish to @compassinformatics npm.

It also removes the need for git submodules, replacing them with `@geoext/geoext` and `@terrestris/basigx` from npm. 
Build config updated to pull in `@geoext/geoext` and `@terrestris/basigx` from `node_modules`
Turf is build postinstall and is included in the npm package so a project like PMS doesn't have to have it's own turf build step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,13 @@ sencha app watch
 
 Open http://localhost:1841 in your browser.
 
+## Override dependencies
+
+To debug against the main branches (or a specific commit) of dependencies, update the `classpath` property in `app.json` to use the `lib` folder instead of `node_modues`, and run the following command to clone git submodules:
+```
+git submodule update --init --recursive
+```
+
 ## Docker install
 
 If a simple local server is needed to omit CORS problems, you can use the one configured in `docker-compose.yml`.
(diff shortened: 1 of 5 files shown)


app.json
```diff
@@ -33,9 +33,9 @@
      */
     "classpath": [
         "app",
-        "./lib/geoext3/src",
```
geographika, line 36: Maybe just comment out these ./lib references. It would be useful to be able to still clone the repos to these locations to debug against the main branches of dependencies. A few times during cpsi-mapview development there were minor fixes that had to be added to GeoExt, and being able to debug and then push a pull request from here would help development.

README.md
```diff
@@ -14,7 +14,6 @@ Clone this repository
 ```
 git clone https://github.com/compassinformatics/cpsi-mapview.git
 cd cpsi-mapview
-git submodule update --init --recursive
```
geographika, line 17: Add note about cloning dependencies to the lib folder if fixes need to be added to the dependencies (see note below).

fergaldoyle:
@geographika  Git Submodules brought back in latest commit along with updated readme

geographika:
@fergaldoyle - thanks. Is this ready to merge now?

fergaldoyle:
@geographika   Yes. After it is merged, next steps would be to publish it manually to npm, and update PMS to use npm instead of git clone.

Then at some stage Github actions could be used to publish new releases to npm automatically on creation of a release on Github.

PR #650, opened 2023-08-16
Add Github action to publish to npm

fergaldoyle:
This PR adds an action to publish to NPM on each Github release.

A secret with name NPM_TOKEN needs to be added under Settings > Secrets & Variables > Actions > New repository secret
The value of NPM_TOKEN is stored in LastPass, PMS shared folder (or a new token can be created under your account on npm)

Just before a release is created, the version number in package.json needs to be incremented.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,19 @@
+name: Publish to NPM
+on:
+  release:
+    types: [created]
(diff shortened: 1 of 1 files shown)


geographika:
@fergaldoyle - the `NPM_TOKEN` has been added to repository secrets.

PR #668, opened 2023-10-11
Match only exact grid xtypes and not Inherited xtypes

geographika:
See https://docs.sencha.com/extjs/7.7.0/classic/Ext.ComponentQuery.html:

To match only the exact type, pass the "shallow" flag by adding (true) to xtype (See Component's [isXType](https://docs.sencha.com/extjs/7.7.0/classic/Ext.Component.html#method-isXType) method):

`prevTextField = myField.previousNode('textfield(true)');`

This allows us to subclass grids and allow them to be associated with a layer - otherwise only one grid will be shown for both layers.


geographika:
Replaced with #673

PR #669, opened 2023-10-12
add the WFS typeName as the fileName

select-8:
For export functionality, the fileName can be passed to the OUTPUTFORMAT parameter as a runtime substitution. The corresponding block in a MapFile can be configured as so:
```
    OUTPUTFORMAT
        NAME "xlsx"
        DRIVER "OGR/XLSX"
        MIMETYPE "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet; charset=utf-8"
        FORMATOPTION "FORM=zip"
        FORMATOPTION "STORAGE=memory"
        FORMATOPTION "FILENAME=%filename%.xlsx"
    END
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ bootstrap.jsonp
 
 lib/*/
 lib/turf.js
+.sencha/ide/config.json
+.sencha/.gitignore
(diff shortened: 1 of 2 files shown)


geographika:
@fergaldoyle  - to be backported to v1.3.0 and a new v1.3.0.1 release to be created.

PR #697, opened 2025-04-08
Allow setting absolute path in groupEditService

fergaldoyle:
This PR allows overriding the serviceUrl generated from `groupEditService` property

If the property is set as is in existing codebases, the `groupEditService` value is appended to viewModel `serviceUrl` value (current behaviour)
```js
{
    groupEditable: true,
    groupEditService: 'isgreenway',
    groupEditDataProp: 'isGreenway',
}
```

If the property starts with a forward slash, `groupEditService` value is used as is, without appending it to viewModel `serviceUrl`
```js
{
    groupEditable: true,
    groupEditService: '/WebServices/namespace/override',
    groupEditDataProp: 'isGreenway',
}
```


select-8:
An override at the column level for idProperty may also be required. For the immediate task, the idProperty will probably be the same for both services, but that might not always be true.

PR #704, opened 2025-07-29
Added management for unfound layers

straatvark:
Mistakes (e.g. duplicate layer keys in the default layer *.json file) result in no layer found. That trigger an exception that stop some layer names to be rendered in the layertree. This pull request manage the exception and give feedback on what the configuration error might be.


geographika:
@straatvark - I modified some of the text for clarity. Now merged.

